Issue #2, opened 2018-07-19
Move verilator test infra from magma to fault


rsetaluri:
magma/magma/testing/verilator.py should be in fault, IMO.

leonardt:
Yup, this is WIP in my changes to refactor the magma_cgra tests to abstract out the compile and run for verilator. The interface should be specifying the target/backend to the Tester, then it should handle all the code for that specific simulator.

leonardt:
If we merge in #7 this will be resolved. Any other comments on that PR? I'm going to spend some time documenting and cleaning up the code base at this point before new features are added. Will submit a PR with that.

leonardt:
Done!

PR #6, opened 2018-07-19
Migrate magma verilator code into fault

leonardt:
Addresses https://github.com/leonardt/fault/issues/2

This was a one-to-one copy, we should review and refactor this code, but unfortunately my nested array support changes depend on changes to this code, so I'd prefer that we refactor after I've merged those changes in.

### DIFF
--- a/fault/tester.py
+++ b/fault/tester.py
@@ -60,7 +60,7 @@ class Tester:
 
     def compile_and_run(self, target="verilator", **kwargs):
         if target == "verilator":
-            target = VerilogTarget(self.circuit, self.test_vectors, **kwargs)
+            target = VerilatorTarget(self.circuit, self.test_vectors, **kwargs)
         else:
             raise NotImplementedError(target)
 
--- a/fault/verilator_target.py
+++ b/fault/verilator_target.py
@@ -11,12 +11,7 @@ def harness(circuit, tests):
 
     assert len(circuit.interface.ports.keys()) == len(tests[0])
 
-    test_vector_length = 0
-    for item in tests[0]:
-        if isinstance(item, Array):
-            test_vector_length += item.flattened_length
-        else:
-            test_vector_length += 1
+    test_vector_length = len(tests[0])
 
     source = '''\
 #include "V{name}.h"
@@ -49,7 +44,6 @@ int main(int argc, char **argv, char **env) {{
 '''.format(len(tests), test_vector_length)
 
     for test in tests:
-        testvector = []
 
         def to_string(t):
             if t is None or t._value is None:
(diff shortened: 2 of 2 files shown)


rsetaluri:
No worries, that works for me. What are the build failures?

fault/tester.py
```diff
@@ -60,7 +60,7 @@ def step(self):
 
     def compile_and_run(self, target="verilator", **kwargs):
         if target == "verilator":
-            target = VerilogTarget(self.circuit, self.test_vectors, **kwargs)
+            target = VerilatorTarget(self.circuit, self.test_vectors, **kwargs)
```
rsetaluri: why did this change? the class name in the other file seems to still be `VerilatorTarget`

fault/verilator_target.py
```diff
@@ -1,5 +1,154 @@
...
+    assert len(circuit.interface.ports.keys()) == len(tests[0])
+
+    test_vector_length = 0
+    for item in tests[0]:
+        if isinstance(item, Array):
+            test_vector_length += item.flattened_length
+        else:
+            test_vector_length += 1
+
+    source = '''\
```
rsetaluri: Eventually we should use a templating engine and move this into a separate template file.

leonardt:
import of the Array module snuck in, that's coming in the next set of changes. just removed it for now.

fault/verilator_target.py
```diff
@@ -1,5 +1,154 @@
...
+        cwd=build_dir, shell=True
+    )
+    assert not subprocess.call(
+        f'make -C obj_dir -j -f V{top_module}.mk V{top_module}', cwd=build_dir,
+        shell=True)
+    assert not subprocess.call('./obj_dir/V{}'.format(top_module),
+                               cwd=build_dir, shell=True)
 
 
 class VerilatorTarget(Target):
```
rsetaluri: Also eventually: we should add a flag as to whether or not the files should be kept around or not. If so, we can put temp files in a dir the user passes in, otherwise can use a temporary dir. The interface could be that if `directory` is None, we put in a tempdir, otherwise in `directory`

fault/verilator_target.py
```diff
@@ -9,9 +158,9 @@ def __init__(self, circuit, test_vectors, directory="build/", flags=[]):
         self._flags = flags
 
     def run(self):
-        magma.testing.verilator.compile(
```
rsetaluri: Also eventually: should some check to ensure the verilog file is here. If not we should compile `circuit` to verilog.

fault/tester.py
leonardt: this was a typo in the original merge, it should be `VerilatorTarget`, it was `VerilogTarget`

rsetaluri: 👍

PR #7, opened 2018-07-19
Nested arrays

leonardt:
Adds support for nested arrays in the IO. Adds two tests, one for the testvector construction, another for running a simple feedthrough module in verilator.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ install:
     # - wget https://raw.githubusercontent.com/phanrahan/magma/master/.travis/install_coreir.sh
     # - source install_coreir.sh
 
+    # FIXME: Workaround until coreir bug where top module is missing is fixed
     - wget https://github.com/rdaly525/coreir/releases/download/v0.0.12/coreir.tar.gz
     - mkdir coreir_release;
     - tar -xf coreir.tar.gz -C coreir_release --strip-components 1;
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,15 +7,9 @@ python:
     - "3.6"
 
 install:
-    # - wget https://raw.githubusercontent.com/phanrahan/magma/master/.travis/install_coreir.sh
-    # - source install_coreir.sh
+    - wget https://raw.githubusercontent.com/phanrahan/magma/master/.travis/install_coreir.sh
+    - source install_coreir.sh
 
-    # FIXME: Workaround until coreir bug where top module is missing is fixed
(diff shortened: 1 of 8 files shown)


.travis.yml
```diff
@@ -7,8 +7,14 @@ python:
     - "3.6"
 
 install:
-    - wget https://raw.githubusercontent.com/phanrahan/magma/master/.travis/install_coreir.sh
```
rsetaluri: can you just note as a comment in this script why this is happening? in case it doesn't get fixed soon 🙂

fault/array.py
```diff
@@ -0,0 +1,28 @@
...
+    @property
+    def flattened_length(self):
+        if isinstance(self.value, Array):
+            return self.N * self.value.flattened_length
+        else:
+            return self.N
+
+    def flattened(self):
+        if isinstance(self.value, Array):
+            return self.value.flattened
```
rsetaluri: should this be a function call?

fault/tester.py
```diff
@@ -27,27 +29,58 @@ def __init__(self, circuit, clock=None):
...
+        elif isinstance(arr, m.ArrayType) and isinstance(arr.T, m.ArrayKind):
+            val = Array([self.get_array_val(x) for x in arr], len(arr))
+        else:
+            raise NotImplementedError(arr, type(arr), arr.T)
+        return val
 
     def get_index(self, port):
         return self.port_index_mapping[port]
 
+    def add_test_vector(self, port, value):
```
rsetaluri: I like this!!

fault/verilator_target.py
```diff
@@ -112,6 +167,7 @@ def compile_verilator_harness(basename, circuit, tests, input_ranges=None):
     if callable(tests):
         tests = testvectors(circuit, tests, input_ranges)
     verilatorcpp = harness(circuit, tests)
+    print(verilatorcpp)
```
rsetaluri: is this a debug artifact?

setup.cfg
```diff
@@ -1,4 +1,4 @@
-[pytest]
+[tool:pytest]
```
rsetaluri: 👍

fault/array.py
leonardt: Fixed, used to be a property, but then I thought it made more sense as a function call

fault/verilator_target.py
leonardt: Yes, removed

.travis.yml
leonardt: Done

rsetaluri: Now we don't need this right?

PR #10, opened 2018-07-23
Add logging module

rsetaluri:
Similar to magma logging module (magma/magma/logging.py).

### DIFF
--- a/fault/logging.py
+++ b/fault/logging.py
@@ -1,7 +1,3 @@
-"""
-Note(rsetaluri): This file is taken almost verbatim from magma/magma/logging.py.
-"""
-
 from __future__ import absolute_import
 from __future__ import print_function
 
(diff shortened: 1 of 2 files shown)


leonardt:
Rebased this branch on master with the pep8 changes merged.

fault/logging.py
```diff
@@ -0,0 +1,50 @@
...
+        else:
+            function = frame.function
+        if function not in ["wire", "connect",
+                            "get_original_wire_call_stack_frame",
+                            "error", "warn"]:
+            break
+    if sys.version_info < (3, 5):
+        return frame[0]
+    else:
+        return frame.frame
```
leonardt: I don't think we need this extra logic for wire tracebacks. Can we remove this function and simplify `error` to basically just printout to stderr? we could just have it use `log.error` like the other functions.

PR #18, opened 2018-08-02
Update implementation for int_vector branch of bit_vector

leonardt:


leonardt:
Tracking https://github.com/leonardt/bit_vector/pull/8

fault/python_simulator_target.py
```diff
@@ -30,7 +31,20 @@ def __check_value(self, port, expected_val):
                             f"is {expected_val}")
         sim_val = self._simulator.get_value(port)
         expected_val = convert_value(expected_val)
-        assert sim_val == expected_val
+        assert self.__check(sim_val, expected_val), \
+            f"Expected {expected_val}, got {sim_val}"
+
+    def __check(self, sim_val, expected_val):
+        if expected_val is None:
+            # Expected None, skipping
```
rsetaluri: should we print some warning in this case? why would expected_val be None?

phanrahan: I believe we are using None to mean any value. This seems likely to cause problems in the future.

leonardt: Perhaps we could use some object called Any? Basically, we need a way of specifying in a testvector to not check an output value. The idea with None is if that "There is no expected value". An Any object could also work: it can be any value.

rsetaluri: Seems like a good idea. As long as it doesn’t confuse ppl with the generic type construct Any. Perhaps AnyValue?

leonardt: Updated to use AnyValue

Issue #16, opened 2018-07-25
Add a way to run a tester on multiple circuit implementations w the same interface


rsetaluri:
Often we want to run a tester on multiple implementations of the "same circuit". For example if we have magma circuits, `foo_impl1` and `foo_impl2` where `foo_impl1.interface == foo_impl2.interface`, we want to create a tester, build test vectors and then run the tests against each implementation. Right now we might do `tester = fault.Tester(foo_impl1)` and then change the circuit later (i.e. `tester.circuit = foo_impl2`) to test the other implementation.

We should build a mechanism where we pass in an interface and then can run the tests against any circuit of that interface.

leonardt:
One option is to separate the notion of test vector generation and running tests. Right now the tester is responsible for both of these. We could pass an interface to a test vector constructor, then have a tester that can rum a test with the test vectors for any circuit that satisfies the interface.

For documentation, here is an example of where this would be used (which currently uses the `tester.circuit` hack): https://github.com/rsetaluri/magma_cgra/blob/master/test_cb/test_cb_regression.py#L115-L118

PR #24, opened 2018-09-11
Add support for copying a tester for a circuit with the same interface

leonardt:
Closes https://github.com/leonardt/fault/issues/16

Basically, we can't just copy test vectors any more because the actions save the associated port. My proposed solution is to add a `copy` method to a tester object, which you can call with a new circuit and clock. It copies all the actions and updates the port/clock references. It also checks that the interfaces are the same.

### DIFF
--- a/fault/tester.py
+++ b/fault/tester.py
@@ -5,6 +5,8 @@ from fault.logging import warning
 from fault.vector_builder import VectorBuilder
 from fault.value_utils import make_value
 from fault.verilator_target import VerilatorTarget
+from fault.actions import Poke, Expect, Step
+import copy
 
 
 class Tester:
@@ -20,11 +22,13 @@ class Tester:
             return VerilatorTarget(self.circuit, self.actions, **kwargs)
         if target == "coreir":
             return MagmaSimulatorTarget(self.circuit, self.actions,
-                                        backend='coreir', **kwargs)
+                                        clock=self.clock, backend='coreir',
+                                        **kwargs)
         if target == "python":
             warning("Python simulator is not actively supported")
             return MagmaSimulatorTarget(self.circuit, self.actions,
-                                        backend='python', **kwargs)
+                                        clock=self.clock, backend='python',
+                                        **kwargs)
         raise NotImplementedError(target)
 
     def poke(self, port, value):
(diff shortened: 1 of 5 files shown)


fault/tester.py
```diff
@@ -53,3 +57,31 @@ def serialize(self):
     def compile_and_run(self, target="verilator", **kwargs):
         target_inst = self.make_target(target, **kwargs)
         target_inst.run()
+
+    def __check_interfaces_match(self, old_circuit, new_circuit):
```
rsetaluri: Seems that this function can be moved outside of the class. Perhaps in common or in this file outside of the class. Maybe rename `check_circuit_is_subset` or something, because we only care that all ports in the old_circuit are in the new circuit.

It maybe also possible to make this late-binding. We may not have a strict subset but still be able to 'copy' because all the accessed ports exist in the new circuit. Depends on how tightly we want to enforce things.

fault/tester.py
```diff
@@ -53,3 +57,31 @@ def serialize(self):
...
+        old_port_names = old_circuit.interface.ports.keys()
+        for name in old_port_names:
+            if not hasattr(new_circuit, name):
```
rsetaluri: More of a style thing, but I've being moving away from using `hasattr` in favor of `name in new_circuit.interface.ports`. I think it is cleaner and more robust. Same throughout this change.

fault/tester.py
```diff
@@ -53,3 +57,31 @@ def serialize(self):
...
+                raise ValueError(f"New circuit does not have port {name}")
+            old_kind = type(type(getattr(old_circuit, name)))
+            new_kind = type(type(getattr(new_circuit, name)))
+            if not (issubclass(new_kind, old_kind) or
+                    issubclass(old_kind, new_kind)):
+                raise ValueError("Types don't match")
+
+    def copy(self, new_circuit, clock=None):
```
rsetaluri: maybe rename to 'retarget'? or something similar. To me copy implies 1-to-1

fault/tester.py
```diff
@@ -53,3 +57,31 @@ def serialize(self):
...
+        """
+        Generates a copy of the tester for new_circuit
+        Checks that new_circut has exactly the same interface as self.circuit
+        """
+        self.__check_interfaces_match(self.circuit, new_circuit)
+        new_tester = Tester(new_circuit, clock)
+        for old_action in self.actions:
+            new_action = copy.copy(old_action)
```
rsetaluri: Maybe actions are required to implement a `retarget(new_circuit)` function? So we can just defer to those calls and don't need the `isinstance()` calls. Both are ok, the former seems cleaner to me -- also depends on if we expect that `retarget` function to be used anywhere else.

tests/test_tester.py
```diff
@@ -46,3 +46,31 @@ def test_tester_nested_arrays():
         expected.append(Expect(circ.O[i], val))
     for i, exp in enumerate(expected):
         check(tester.actions[i], exp)
+
```
rsetaluri: Good test!

leonardt:
Changed to use the `retarget` pattern as suggested. The `Action` ABC now includes an abstract method `retarget(self, new_circuit, clock)`. We have to pass the clock for the step action, so the other actions ignore that argument (also Eval ignores all the arguments).

rsetaluri:
Looks good. Made some minor refactoring changes.

PR #27, opened 2018-09-11
Add a mechanism for including verilog in Verilator target

leonardt:
Basically, I need to include the cells defined in https://raw.githubusercontent.com/YosysHQ/yosys/master/techlibs/ice40/cells_sim.v so I could use the verilator target to simulate output from mantle40 (which references those SB_ primitives).

Here's the quick way to inject the verilog, any ideas on a better interface? Right now the verilator target accepts a string which will be prepended to the compiled file from magma. This is sufficient to implement what I need.


tests/test_include_verilog.py
```diff
@@ -0,0 +1,25 @@
+import fault
+import magma as m
+
+
+def test_include_verilog():
+    with open("tests/cells_sim.v") as f:
```
rsetaluri: probably should test this feature with a different/simple file, to avoid including a big verilog file.

fault/verilator_target.py
```diff
@@ -102,6 +103,12 @@ def run(self):
             prefix = str(verilog_file)[:-2]
             magma.compile(prefix, self.circuit, output="verilog")
         assert verilog_file.is_file()
+        if self.include_verilog:
```
rsetaluri: does verilator allow passing multiple file names? then we can just pass all the filenames without modifying this one. As long as we properly specify the top, it seems like this would work.

rsetaluri: I believe multiple files passed to verilator does indeed work as expected.

leonardt:
Okay, trimmed the .v file to just include the SB_DFF definition which is used in the test.

I also updated the interface to use the `-v` argument for verilator which allows the including of verilog libraries. In this case, you need to pass the path to the files relative to where the verilator command is invoked.

rsetaluri:
Looks good. Just made a commit to use temporary directory instead of checking a build directory.

PR #28, opened 2018-09-14
Add initial support for print

leonardt:
Only works on top level ports

Supports an optional argument `format_str` which allows you to customize how the value is printed (by default it is `%x`).

Example usage:
```python
    tester.print(circ.O, "%08x")
```

Adds tests for Actions, Tester, and the magma/verilator targets.

### DIFF
--- a/fault/actions.py
+++ b/fault/actions.py
@@ -52,7 +52,6 @@ class Peek(Action):
         return cls(new_port, self.format_str)
 
 
-
 class Expect(PortAction):
     def __init__(self, port, value):
         if port.isoutput():
--- a/fault/actions.py
+++ b/fault/actions.py
@@ -34,6 +34,25 @@ class Poke(PortAction):
         super().__init__(port, value)
 
 
+class Peek(Action):
+    def __init__(self, port, format_str="%x"):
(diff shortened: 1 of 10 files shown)


fault/actions.py
```diff
@@ -34,6 +34,24 @@ def __init__(self, port, value):
         super().__init__(port, value)
 
 
+class Peek(Action):
```
rsetaluri: Wondering if we can rename this to like `Print(format_str, *args)` where we have some special syntax for a port (like `%v` or something else that's not used). The target functionality is the same, it just makes it more clear that this is just a print statement, not returning you the value inline as a `peek` normally would.

I.e. `val = peek(foo.O)` doesn't make sense here, though that's the normal usage of peek. Then if we have a non-compiled tester in the future that's running in real time, we can actually have a `peek` that returns a value.

leonardt: Okay that makes sense.

leonardt:
If you're referring to the tester method `peek`, just caught that (sorry was find/replacing Peek, didn't catch the lower case)

rsetaluri:
oh just saw you did that 😄

PR #29, opened 2018-09-16
[verilator-target] Add interface to add include directories

leonardt:
Adds a parameter `include_directories` which is a list of directories to
include in the verilator command using the `-I` flag.

Also, renames the include_verilog_files parameter/field to be
`include_verilog_libraries` to more explicitly match the verilator
terminology (`-v` is use for library files.)

### DIFF
--- a/fault/verilator_target.py
+++ b/fault/verilator_target.py
@@ -37,12 +37,14 @@ int main(int argc, char **argv) {{
 
 class VerilatorTarget(Target):
     def __init__(self, circuit, actions, directory="build/",
-                 flags=[], skip_compile=False, include_verilog_files=[]):
+                 flags=[], skip_compile=False, include_verilog_libraries=[],
+                 include_directories=[]):
         super().__init__(circuit, actions)
         self.directory = Path(directory)
         self.flags = flags
         self.skip_compile = skip_compile
-        self.include_verilog_files = include_verilog_files
+        self.include_verilog_libraries = include_verilog_libraries
+        self.include_directories = include_directories
 
     @staticmethod
     def generate_action_code(i, action):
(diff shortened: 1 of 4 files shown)


fault/verilator_target.py
```diff
@@ -37,13 +37,14 @@
 
 class VerilatorTarget(Target):
     def __init__(self, circuit, actions, directory="build/",
-                 flags=[], skip_compile=False, include_verilog_files=[],
-                 magma_output="verilog"):
+                 flags=[], skip_compile=False, include_verilog_libraries=[],
+                 include_directories=[], magma_output="verilog"):
```
rsetaluri: What is the use case for this flag? Is there a simple test case for it?

A small note documenting the difference between these 2 flags would be great.

leonardt: It allows your specify a directory that contains verilog files you want to include.

`include_verilog_libraries` corresponds to the `-v` flag in verilator
`include_directories` corresponds to the `-I` flag in verilator

I think the functionality is pretty much the same, but it's convenient to provide both options depending on what the user has ready (a file name or a directory path). That's what I could glean from the verilator help string from https://www.veripool.org/projects/verilator/wiki/Manual-verilator

```
     -I<dir>                    Directory to search for includes
...
     -v <filename>              Verilog library
...
```

leonardt: One issue is that when using `-I`, verilator searches for verilog files with the same name as the module it's trying to include. When you pass a file `-v`, verilator loads all the modules in that file into the namespace, so the semantics are slightly different (e.g. you can't use -I if your module name doesn't match the file name it's contained in).

leonardt: I suppose you could achieve this using the `flags` parameter, so this is really a convenience method for not having to build the `-I` strings, and instead you can just pass a list of directories.

leonardt: Added a test

rsetaluri: Sounds good to me. Thanks for the test.

PR #30, opened 2018-09-16
Allow printing input values

leonardt:
Can be useful for debugging to see what the inputs are at the time of
expect failure.

Also, the error message has a stray `peek`

### DIFF
--- a/fault/actions.py
+++ b/fault/actions.py
@@ -36,9 +36,6 @@ class Poke(PortAction):
 
 class Print(Action):
     def __init__(self, port, format_str="%x"):
-        if port.isoutput():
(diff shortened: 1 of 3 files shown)


fault/actions.py
```diff
@@ -36,9 +36,6 @@ def __init__(self, port, value):
 
 class Print(Action):
     def __init__(self, port, format_str="%x"):
-        if port.isoutput():
```
rsetaluri: Does this work for magma, i.e. does `simulator.get_value(port)` work for input ports? I can add a test for this.

rsetaluri: Added test. It works.

PR #32, opened 2018-09-16
Rename vector generator to action generator

leonardt:
This is just a name change to better reflect the semantics of the tester API. Since the action API is default, I think we should have these generators create actions, then the user can use the `serialize` method to get test vectors.

### DIFF
--- a/fault/action_generators.py
+++ b/fault/action_generators.py
@@ -1,0 +1,31 @@
+import fault
+from fault.common import get_renamed_port
+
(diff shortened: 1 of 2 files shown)


fault/action_generators.py
```diff
@@ -0,0 +1,32 @@
+import fault
+from fault.common import get_renamed_port
+
+
+def generate_actions_from_streams(circuit, functional_model, streams,
```
rsetaluri: Do we plan to do the same for the function `generate_random_test_vectors`?

PR #36, opened 2018-09-17
Add support for poke/expect using a list of values for nested arrays

leonardt:
Support passing a list of values as an argument to `poke` and `expect` for nested arrays. This still supports passing a single value, which is convenient if you want to zero all the children of the nested arrays, so you can say something like `tester.poke(circuit.nested_array, 0)`. 

Also includes a default `__repr__` method for actions that uses `__str__` for now (useful for debugging when printing a list of actions, which calls repr)

### DIFF
--- a/fault/magma_simulator_target.py
+++ b/fault/magma_simulator_target.py
@@ -27,7 +27,10 @@ class MagmaSimulatorTarget(Target):
                 isinstance(port.T, m._BitType) and \
                 not isinstance(port, m.BitsType) and \
                 isinstance(expected, BitVector):
-            # Python simulator will return a list of bools
+            # If port is an Array(N, Bit) and **not** a Bits(N), then the
+            # Python simulator will return a list of bools. So, if the user
+            # provided a BitVector, we unpack it here so the equality check
+            # works
             expected = expected.as_bool_list()
         if isinstance(port, m.ArrayType):
             for i in range(port.N):
--- a/fault/magma_simulator_target.py
+++ b/fault/magma_simulator_target.py
@@ -51,7 +51,7 @@ class MagmaSimulatorTarget(Target):
                 if isinstance(action.port, (m.ArrayType, m.ArrayKind)) and \
                         isinstance(action.port.T, (m._BitType, m._BitKind)):
                     got = BitVector(got).as_uint()
-                else if isinstance(action.port, (m.ArrayType, m.ArrayKind)):
+                elif isinstance(action.port, (m.ArrayType, m.ArrayKind)):
                     raise NotImplementedError("Printing complex nested arrays")
                 print(f'{action.port.debug_name} = {action.format_str}' %
                       got)
--- a/fault/magma_simulator_target.py
+++ b/fault/magma_simulator_target.py
@@ -23,9 +23,9 @@ class MagmaSimulatorTarget(Target):
 
     @staticmethod
     def check(got, port, expected):
-        if isinstance(port, (m.ArrayType, m.ArrayKind)) and \
-                isinstance(port.T, (m._BitType, m._BitKind)) and \
-                not isinstance(port, (m.BitsType, m.BitsKind)):
+        if isinstance(port, m.ArrayType) and \
+                isinstance(port.T, m._BitType) and \
+                not isinstance(port, m.BitsType):
             if isinstance(expected, BitVector):
                 # Python simulator will return a list of bools
                 expected = expected.as_bool_list()
@@ -48,10 +48,10 @@ class MagmaSimulatorTarget(Target):
                 simulator.set_value(action.port, value)
             elif isinstance(action, actions.Print):
                 got = simulator.get_value(action.port)
-                if isinstance(action.port, (m.ArrayType, m.ArrayKind)) and \
+                if isinstance(action.port, m.ArrayType) and \
                         isinstance(action.port.T, (m._BitType, m._BitKind)):
                     got = BitVector(got).as_uint()
-                elif isinstance(action.port, (m.ArrayType, m.ArrayKind)):
+                elif isinstance(action.port, m.ArrayType):
                     raise NotImplementedError("Printing complex nested arrays")
                 print(f'{action.port.debug_name} = {action.format_str}' %
                       got)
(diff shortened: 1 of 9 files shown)


fault/tester.py
```diff
@@ -1,4 +1,5 @@
 import magma
+import magma as m
```
rsetaluri: Let's just use one of these imports; impartial to if we use `as m` or not.

fault/tester.py
```diff
@@ -33,18 +34,33 @@ def make_target(self, target, **kwargs):
                                         **kwargs)
         raise NotImplementedError(target)
 
+    def make_array_action(self, port, value, action):
+        for i in range(port.N):
+            action(port[i], value[i])
+
     def poke(self, port, value):
         value = make_value(port, value)
-        self.actions.append(actions.Poke(port, value))
```
rsetaluri: I think this logic should be delegated to the targets -- tester should ideally not be implementing logic to split a `poke(array_port, [...])` into several actions. I would argue these functions should be unchanged, and in the targets we do a:

```python
if isinstance(value, fault.Array):
   ...
```

Because, the magma/coreir simulator may support directly `set_value(array_port, [...])` where as of course for verilator we need to blow it up.

tests/test_tester.py
```diff
@@ -52,6 +52,22 @@ def test_tester_nested_arrays():
         check(tester.actions[i], exp)
 
 
+def test_tester_nested_arrays_list_of_vals():
```
rsetaluri: Can we copy this test for the actual targets and ensure that it works?

fault/tester.py
leonardt: Good point, that's actually how I initially implemented it. I'll make this change.

leonardt:
Review comments addressed, ready for another pass.

The logic has been moved to the targets.

fault/magma_simulator_target.py
```diff
@@ -33,13 +47,17 @@ def run(self):
                     value = value.as_uint()
                 simulator.set_value(action.port, value)
             elif isinstance(action, actions.Print):
-                got = BitVector(simulator.get_value(action.port))
+                got = simulator.get_value(action.port)
+                if isinstance(action.port, (m.ArrayType, m.ArrayKind)) and \
+                        isinstance(action.port.T, (m._BitType, m._BitKind)):
+                    got = BitVector(got).as_uint()
+                else if isinstance(action.port, (m.ArrayType, m.ArrayKind)):
+                    raise NotImplementedError("Printing complex nested arrays")
```
leonardt: I think we should defer this feature for now. The user can print elements of the array fine, but there's not a simple way to map a complex nested array to a standard format string.

fault/verilator_target.py
```diff
@@ -46,9 +52,20 @@ def __init__(self, circuit, actions, directory="build/",
         self.include_verilog_files = include_verilog_files
         self.magma_output = magma_output
 
+    @staticmethod
+    def make_array_action_code(i, action):
```
rsetaluri: nit: `generate` rather than `make`

rsetaluri: Handy function! But it seems it only handles one level of nesting. Perhaps we could throw the multi-level flattening into `flatten`? Or explicitly error if we don't want to support it (though I don't see any reason it wouldn't work here).

fault/magma_simulator_target.py
```diff
@@ -33,13 +47,17 @@ def run(self):
                     value = value.as_uint()
                 simulator.set_value(action.port, value)
             elif isinstance(action, actions.Print):
-                got = BitVector(simulator.get_value(action.port))
+                got = simulator.get_value(action.port)
+                if isinstance(action.port, (m.ArrayType, m.ArrayKind)) and \
```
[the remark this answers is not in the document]
leonardt: This PR only addresses arrays, which is blocking https://github.com/StanfordAHA/garnet/pull/105#issuecomment-422091599, can we save tuple support for separate PR?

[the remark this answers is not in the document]
leonardt: Not sure about Type and Kind, let me look into it, I think I saw other code that was like this so I was just duplicating the Type Kind check

fault/magma_simulator_target.py
```diff
@@ -33,13 +47,17 @@ def run(self):
...
+                        isinstance(action.port.T, (m._BitType, m._BitKind)):
+                    got = BitVector(got).as_uint()
+                elif isinstance(action.port, (m.ArrayType, m.ArrayKind)):
```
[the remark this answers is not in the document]
leonardt: The issue here for printing complex nested arrays is that it doesn't map to our `format` string parameter. Copied my comment below that seemed to get lost in this change set, maybe from the rebase.

> I think we should defer this feature for now. The user can print elements of the array fine, but there's not a simple way to map a complex nested array to a standard format string.

fault/verilator_target.py
leonardt: it handles multiple levels of nesting, because it recursively calls `generate_action_code` which, if it encounters a nested array, will return a flattened list, so our `flatten` only needs to handle one level of flattening (assuming that everything is recursively flattened).

leonardt: e439fa3 tests this explicitly

fault/magma_simulator_target.py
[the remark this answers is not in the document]
leonardt: Actually, it's not just `isinstance(action.port, (m.BitsType, m.BitsKind))`, because 
```
>>> a = m.Array(2, m.Bit)
>>> isinstance(a, m.BitsKind)
False
```
This allows you to print a simple array of bits (that wasn't explicitly a bits type)

fault/magma_simulator_target.py
```diff
@@ -21,6 +21,20 @@ def simulator_cls(backend):
             return PythonSimulator
         raise NotImplementedError(backend)
 
+    @staticmethod
+    def check(got, port, expected):
```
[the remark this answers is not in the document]
rsetaluri: Got it

fault/verilator_target.py
rsetaluri: Ah, right, good point.

rsetaluri:
Looks good, thanks for addressing the comments.

fault/magma_simulator_target.py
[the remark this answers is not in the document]
rsetaluri: Yeah, I think I would like to implement a more general thing, like `Print(format_string, *args)` where we have special syntax for single ports, arrays, and tuples. Think it can be deferred for future work. This is good for now.

PR #46, opened 2018-10-09
verilator --trace

THofstee:


fault/verilator_target.py
```diff
@@ -34,8 +41,21 @@ def flatten(l):
   Verilated::commandArgs(argc, argv);
   V{circuit_name}* top = new V{circuit_name};
 
+#if VM_TRACE
+  vluint64_t main_time = 0;
+
+  Verilated::traceEverOn(true);
+  tracer = new VerilatedVcdC;
+  top->trace(tracer, 99);
+  Verilated::mkdir("logs");
```
leonardt: This didn't work with `verilator 3.926`, which is the current released version of homebrew, but it does with `brew install verilator --HEAD`, so it must be a new feature. Could we change this to use a C++ stdlib function to maintain compatibility? Small change and this is a broadly useful feature so it would be good to not have to require people to be on a bleeding edge version (especially in industrial systems running an old OS)

fault/verilator_target.py
```diff
@@ -34,8 +41,21 @@ def flatten(l):
...
+  tracer->open("logs/vlt_dump.vcd");
```
leonardt: Also, let's change this to use the circuit name or verilog file name instead of the generic one, so we can distinguish between traces from different files.

PR #50, opened 2018-12-05
Initialize System Verilog Target

leonardt:
This PR adds initial support for commercial simulators by generating a SystemVerilog test bench.

It's been tested locally using NCSIM, before merging I'd also like to test and use VCS.

@rsetaluri There's some refactoring to do here since there's a lot of shared logic between the verilator target and the system verilog target. One nice thing is that we can use the same test suite, and parametrize over target. This means the new SV backend supports all the tests (and use cases) of the verilator backend, so any existing verilator tests should in theory be portable to this new target.

I'd also like to add some end to end tests (right now it instantiates the target directly, but we should be testing the code paths for users via the `Tester` object). One thing we should test is the ability to invoke different simulators (NCSIM vs VCS).

@rsetaluri can you review these changes with those notes in mind? @alexcarsello and @priyanka-raina have been asking for this feature, so I'd like to get the changes merged ASAP so they can start testing it out. This shouldn't affect any existing test benches, so I'm not worried about that, but may need to be careful if I end up refactoring the verilator target to share code with this new target (example the recursive array code is duplicated). So, basically I'm looking for a review (approval) on functionality/design (modulo refactoring code).

### DIFF
--- a/fault/system_verilog_target.py
+++ b/fault/system_verilog_target.py
@@ -1,0 +1,145 @@
+from fault.verilog_target import VerilogTarget
+import magma as m
+from pathlib import Path
+import fault.actions as actions
+import fault.verilator_utils as verilator_utils
+from bit_vector import BitVector
+import fault.value_utils as value_utils
+import subprocess
+
+
+src_tpl = """\
+`timescale 1ns/1ns
+
+module {circuit_name}_tb;
+{declarations}
+    initial begin
+{initial_body}
+        #20 $finish;
+    end
+
+    {circuit_name} dut (
+        {port_list}
+    );
+
+endmodule
+"""
+
+
+class SystemVerilogTarget(VerilogTarget):
+    def __init__(self, circuit, circuit_name=None, directory="build/",
+                 skip_compile=False, magma_output="verilog"):
+        super().__init__(circuit, circuit_name, directory, skip_compile,
+                         magma_output)
+
+    @staticmethod
+    def generate_action_code(i, action):
+        if isinstance(action, actions.Poke):
+            if isinstance(action.port, m.ArrayType) and \
+                    not isinstance(action.port.T, m.BitKind):
+                return VerilatorTarget.generate_array_action_code(i, action)
+            name = verilator_utils.verilator_name(action.port.name)
+            if isinstance(action.value, BitVector) and \
+                    action.value.num_bits > 32:
+                raise NotImplementedError()
+            else:
+                value = action.value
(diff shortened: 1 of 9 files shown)


leonardt:
Also note, the lack of availability of commercial simulators will mean that we can't test this target on travis. I'm working on setting up a BuildKit agent on a local stanford VM so we can try a CI flow with our own infrastructure. This can be a test run of this process, which we can then add to garnet. For now, I think we can rely on the local tests passing so we can ship this out to alex and priyanka asap.

fault/system_verilog_target.py
```diff
@@ -0,0 +1,191 @@
...
+import fault.actions as actions
+import fault.verilator_utils as verilator_utils
+from bit_vector import BitVector
+import fault.value_utils as value_utils
+import subprocess
+from fault.util import flatten
+
+
+src_tpl = """\
+`timescale 1ns/1ns
```
rsetaluri: How important is this parameter? Does this need to be a parameter? Maybe @alexcarsello can help here.

fault/system_verilog_target.py
```diff
@@ -0,0 +1,191 @@
...
+        if isinstance(action, actions.Step):
+            name = verilator_utils.verilator_name(action.clock.name)
+            code = []
+            for step in range(action.steps):
+                # code.append("top->eval();")
+                # code.append("main_time++;")
+                # code.append("#if VM_TRACE")
+                # code.append("tracer->dump(main_time);")
+                # code.append("#endif")
+                code.append(f"#5 {name} ^= 1;")
```
rsetaluri: why `#5`?

fault/system_verilog_target.py
leonardt: We can make `timescale` a parameter to the target, so have this by default for most use cases, but allow the user to set it.

fault/system_verilog_target.py
leonardt: Based on an example from @priyanka-raina https://github.com/priyanka-raina/fault_vcs_test/blob/master/verilog_tester_counter.v

I think this should also be a parameter. Basically, from what I understand, `#` indicates delay, and the number of timesteps to delay depends on your clock period. Using `#5` for each step indicates a clock period of 10 steps. Not sure what a good default is, but I think this should definitely be configurable by the user.

leonardt:
Okay, I've added end-to-end tests (using tester.compile_and_run) and refactored the implementation to share a lot more code (generate code logic is reused, the targets implement specific `make_<action>` functions to translate the action to code.

PR #51, opened 2018-12-14
Sv include verilog

leonardt:
Adds support for the `include_verilog_libraries` parameter to the system verilog target. This allows the user to provide a list of verilog files which will be passed to the simulator when running (e.g. for including standard cell libraries).

This functionality already existed for the verilator target, so I moved the interface to the `VerilogTarget` parent class and updated the test to be parametrized over target.

I have not yet added include directories because I'm not aware of how to do this with ncsim/vcs, but should be simple enough to add if we need it (and can figure out the right option to pass to the simulators).

### DIFF
--- a/fault/system_verilog_target.py
+++ b/fault/system_verilog_target.py
@@ -183,7 +183,7 @@ class SystemVerilogTarget(VerilogTarget):
         with open(test_bench_file, "w") as f:
             f.write(src)
         verilog_libraries = " ".join(str(x) for x in
-                                      self.include_verilog_libraries)
+                                     self.include_verilog_libraries)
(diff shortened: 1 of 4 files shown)


fault/system_verilog_target.py
```diff
@@ -180,18 +181,19 @@ def run(self, actions):
         # Write the verilator driver to file.
         src = self.generate_code(actions)
         with open(test_bench_file, "w") as f:
-            print(src)
             f.write(src)
+        verilog_libraries = " ".join(str(x) for x in
+                                      self.include_verilog_libraries)
```
rsetaluri: heads up that lint is complaining here.

Issue #56, opened 2019-01-23
Add ability to peek internal signals


Kuree:
In many cases it would be useful to allow users to peek internal signals to verify the logic. For instance, in interconnect testing, I'd like to to peek the PE core to see if the signal has been routed correctly.

leonardt:
Closed by #57

PR #57, opened 2019-01-24
Add primitive support for internal signals

leonardt:
Submitting this PR with the intention of blocking @Kuree. Basically, we need the ability to expect/peek internal signals. One issue is how to do this with wrapped verilog.

@Kuree, I think this should give the basic functionality required to expect values on the internal registers of the wrapped PE. You need to use `verilator_include` to include the public module (verilog module), you need to add a comment in the verilog file on the signals you want to be public, and then you can use `WrappedVerilogInternalPort`. It's clunky, but let me know if this gets you the functionality you require. Can you try checking out and installing this branch of fault? There are docs at https://github.com/leonardt/fault/blob/9c978320c970692b051bb4eaa11420f59d9f6910/doc/actions.md#wrappedveriloginternalport, let me know if you have any questions.

I'm going to continue to work on making this interface easier to use (ideally we can automatically determine the require `verilator_includes` and make it easier to specify a wrapped verilog signal by using standard dot notation) as well as extend it to support internal magma signals (this should be simpler to implement) as well as for the system verilog backend targets.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,10 +5,17 @@ addons:
     - ubuntu-toolchain-r-test
     packages:
     - g++-4.9
-    - verilator
 python:
 - '3.6'
 install:
+- sudo apt-get install -y make autoconf g++ flex bison
+- wget https://www.veripool.org/ftp/verilator-4.008.tgz
+- unset VERILATOR_ROOT
+- tar xvzf verilator*.t*gz
+- cd verilator*
+- ./configure
+- make
+- sudo make install
 - wget https://raw.githubusercontent.com/phanrahan/magma/master/.travis/install_coreir.sh
 - source install_coreir.sh
 - pip install python-coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ install:
 - ./configure
 - make
 - sudo make install
+- cd ..
 - wget https://raw.githubusercontent.com/phanrahan/magma/master/.travis/install_coreir.sh
 - source install_coreir.sh
 - pip install python-coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ install:
 - pip install -r requirements.txt
 - pip install -e .
 script:
-- pytest --cov fault --codestyle fault -v --cov-report term-missing tests
+- pytest --cov fault --codestyle fault -s -v --cov-report term-missing tests
 after_success:
 - coveralls
 deploy:
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ install:
 - pip install -r requirements.txt
 - pip install -e .
 script:
-- pytest --cov fault --codestyle fault -s -v --cov-report term-missing tests
+- pytest --cov fault --codestyle fault -v --cov-report term-missing tests
 after_success:
 - coveralls
 deploy:
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,18 +5,10 @@ addons:
     - ubuntu-toolchain-r-test
     packages:
     - g++-4.9
+    - verilator
 python:
 - '3.6'
 install:
-- sudo apt-get install -y make autoconf g++ flex bison
-- wget https://www.veripool.org/ftp/verilator-4.008.tgz
-- unset VERILATOR_ROOT
-- tar xvzf verilator*.t*gz
-- cd verilator*
-- ./configure
-- make
-- sudo make install
-- cd ..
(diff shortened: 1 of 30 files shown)


leonardt:
Begins to address https://github.com/leonardt/fault/issues/56

Kuree:
Thanks. I might need some extra tweaking since I need to wrap the circuit into `Core`.

 I'll be busy with some personal matter later this week. I will try it out as soon as I'm available.

leonardt:
Okay, that sounds like you need support for a hierarchy that includes magma and verilog modules, for example:
```
top
    | magma_core
        | verilog pe_core
             | verilog reg
```

I'll add a test/support for this case.

leonardt:
Update:

I have a basic example working with magma internal signals. It depends on this branch of coreir (https://github.com/rdaly525/coreir/pull/659)

Here's the sample circuit: https://github.com/leonardt/fault/blob/838fb59510e0233d10959d34d655ace6086f8ca1/tests/common.py#L46-L68

Here's the test:
https://github.com/leonardt/fault/blob/838fb59510e0233d10959d34d655ace6086f8ca1/tests/test_setattr_interface.py#L6-L21

Notice that we can expect the output of the internal register with
```python
        tester.circuit.config_reg.Q.expect(i)
```

This test is using the new `setattr`/`getattr` interface, so instead of using `poke` and `expect` methods on the tester, you must use the pattern `tester.circuit.<instance>...<port>` or `tester.circuit.<port>`. You can set the port attribute to perform a poke, you can use the `.expect` method on a port attribute to perform an expect.

Still need to add support for poking an internal signal, but not sure how these semantics work w.r.t. the simulator. That is, we don't have support for leaving internal signals unwired (because of coreir), so this would mean you are poking a signal that has another driver, possibly with another value. In theory, the simulator might be smart enough to only execute downstream computation and update the dependent values (if it's event driven), however because verilator uses `eval` semantics, I suspect that poking an internal signal will have no effect (assuming verilator executes in topological order, the poked internal signal would be overwritten by the driver). Is this a pattern we need to support?

@Kuree can you test this out? Something might break so let me know what issues you run into. I'm going to add some more tests with more complex hierarchies to make sure the nesting logic works. Let me know if you have questions about how to use it.

Kuree:
Thanks! I will try it out tomorrow.

leonardt:
Just found out that you'll also need the latest master/release of magma (`magma-lang` pip package v0.1.8) which adds support for the `verilator_debug` option when compiling magma through coreir. The `verilator_debug` option is also now available in the latest coreir master/release.

Kuree:
@leonardt Probably not the best place to report it. After switching to the latest coreir/magma, I got this error when running pytest
```
ImportError while loading conftest '/home/keyi/workspace/garnet/conftest.py'.
conftest.py:2: in <module>
    from magma.circuit import magma_clear_circuit_cache
E   ImportError: cannot import name 'magma_clear_circuit_cache' from 'magma.circuit' (/home/keyi/workspace/garnet/env/lib/python3.7/site-packages/magma/circuit.py)
```

Here is my `pip freeze` output
```
$ pip freeze
astor==0.7.1
atomicwrites==1.2.1
attrs==18.2.0
bit-vector==0.39a0
colorlog==4.0.2
coreir==0.31a0
fault==0.35
graphviz==0.10.1
Jinja2==2.10
magma-lang==0.1.7
Mako==1.0.7
mantle==0.1.10
MarkupSafe==1.1.0
more-itertools==5.0.0
numpy==1.16.0
ordered-set==3.1
-e git://github.com/phanrahan/pe.git@c50b5f6010f9893e4e60db336f5183991ccb2899#egg=pe
pluggy==0.8.1
py==1.7.0
pytest==4.1.1
-e git://github.com/leonardt/Pyverilog.git@ede3155e89a4fb1bb860d772defac2f8eb291b8c#egg=pyverilog
six==1.12.0
```

leonardt:
Turns out you also need this patch to coreir https://github.com/rdaly525/coreir/pull/660 (I think my original test was passing because I was using an old version of coreir locally)

leonardt:
@Kuree You'll need the changes in this branch: https://github.com/StanfordAHA/garnet/pull/146/files#diff-c9b030b0828c1291b56fc10aba6baab6.
IIRC, there were some other bugs that popped up in our test suite when we upgraded magma versions, so that was blocking this branch from getting merged.

leonardt:
Latest version of this branch depends on https://github.com/rdaly525/coreir/pull/666 to add support for poking internal state registers. This support is preliminary and quite brittle, so we should monitor usage and expect bugs to arise. Will work on adding support for the system verilog targets next (should be simpler).

Issue #61, opened 2019-02-11
Issue with setting nested arrays


rsetaluri:
```RecursionError: maximum recursion depth exceeded while calling a Python object
Exception ignored in: '_pydevd_frame_eval.pydevd_frame_evaluator_darwin_36_64.get_bytecode_while_frame_eval'
RecursionError: maximum recursion depth exceeded while calling a Python object
RecursionError: maximum recursion depth exceeded while calling a Python object
Exception ignored in: '_pydevd_frame_eval.pydevd_frame_evaluator_darwin_36_64.get_bytecode_while_frame_eval'
RecursionError: maximum recursion depth exceeded while calling a Python object
F
tests/helper_test_readyvalid.py:211 (test_clock_adjusted_2dlb_flicker_ce_with_2x2_stride)
def test_clock_adjusted_2dlb_flicker_ce_with_2x2_stride():
        c = coreir.Context()
        cirb = CoreIRBackend(c)
    
        testcircuit = DefineTwoDimensionalLineBuffer(cirb, Array(8, In(Bit)), 1, 1, 2, 2, 8, 8, 2, 2, 0, 0, True)
    
        tester = fault.Tester(testcircuit, testcircuit.CLK)
    
        for i in range(30):
            if i % 2 == 0:
>               tester.circuit.I[0][0] = int2seq(1, 8)
E               TypeError: 'PortWrapper' object does not support indexing

helper_test_readyvalid.py:222: TypeError```

leonardt:
#62 has a preliminary support for setting arrays (see this test https://github.com/leonardt/fault/pull/62/files#diff-be25ab5c75d3019d597f051e308a886cR84), let me know if this fixes the issue.

Can I check out the above test somewhere to try it out? (I'm guessing it's in aetherling, CCing @david-durst)

David-Durst:
https://github.com/David-Durst/aetherling/blob/2043af4641f5bf46f5dd111d0812acf98f4eb932/tests/helper_test_readyvalid.py#L19-L72

David-Durst:
Note, in order to run this, you'll need the david-ram-tests magma, mantle, aetherling and fault branches (at least as of 6 pm this evening)

leonardt:
I had to make the following changes (basically replace `int2seq` with `BitVector`).
```diff
diff --git a/tests/helper_test_readyvalid.py b/tests/helper_test_readyvalid.py
index 723cb17..cf5efe0 100644
[...]

PR #62, opened 2019-02-12
Add support for arrays/tuples to setattr interface

leonardt:
Addresses #61

### DIFF
--- a/fault/wrapper.py
+++ b/fault/wrapper.py
@@ -1,5 +1,6 @@
 import fault
 from fault.select_path import SelectPath
+import magma as m
 
 
 class Wrapper:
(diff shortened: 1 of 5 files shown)


fault/wrapper.py
```diff
@@ -49,11 +50,49 @@ class PortWrapper:
...
+            if not isinstance(self.port, (m.TupleType)):
+                raise Exception(f"Can only use getattr with tuples, "
+                                f"not {type(self.port)}")
+            select_path = self.select_path
+            return PortWrapper(getattr(self.port, key), self.parent)
+        except AttributeError:
+            return object.__getattribute__(self, key)
+
+    def __setattr__(self, key, value):
+        try:
```
rsetaluri: I know it's clunky, but since this is internal and `__init__` is not that big, maybe just using `object.__setattr__()` in `__init__` and avoiding this try-catch here would work? Unless I'm missing something.

rsetaluri: We could also do the same in the other place where we need to do this hack.

leonardt: using setattr would work, but is that really better than this try catch? Both solutions suck, I don't think it's that much better. If we want to add fields to __init__ later, that solution is more annoying. This one works, a metaclass pipeline is probably the right way to do it, but I won't have time to investigate that until after Thursday.

leonardt: I think there were calls to setattr/getattr even before init (in the instance creation pipeline) which was another reason that overriding the default behavior even with using `getattr` in `__init__`, but I don't remember specifically (this is the result of hacking my way through to a working solution). You're welcome to explore an alternative.

rsetaluri: Fair enough, didn't realize the complexity behind that solution too. Fine with this for now.

PR #65, opened 2019-02-27
Coverage

leonardt:
Based on `assume-guarantee` branch, shows minimal example for getting toggle coverage out of verilator.

Line coverage does not seem to produce any interesting result (returns `Total coverage (0/0) 0.00%`, maybe this is supposed to be used with SystemC?

Still investigate support for ncsim/vcs coverage


leonardt:
Closing in favor of #205

PR #84, opened 2019-04-22
Allow Printing To Logs That Can Be Parsed By External Programs

David-Durst:
These changes make the verilator backend:
1. Not print the standard prefix to each line
2. Write the output to a file
3. Print lines with no value

These changes are inconsistent with the other backends. I presume better changes will be consistent across all backends.


leonardt:
Closed in favor of https://github.com/leonardt/fault/pull/88

PR #87, opened 2019-04-25
Constrained random

cdonovick:
require https://github.com/leonardt/hwtypes/pull/23

### DIFF
--- a/fault/random.py
+++ b/fault/random.py
@@ -2,6 +2,12 @@ import random
 from hwtypes import BitVector, Bit
 
 
+def constrained_random_bv(width, pred):
+    while True:
+        randval = random_bv(width)
+        if pred(randval):
+            return randval
+
 def random_bv(width):
     return BitVector[width](random.randint(0, (1 << width) - 1))
 
--- a/fault/random.py
+++ b/fault/random.py
@@ -13,15 +13,8 @@ def constrained_random_bv(width, pred):
         if pred(randval):
             return randval
 
-def constrained_random_bv(width, pred):
-    while True:
-        randval = random_bv(width)
-        if pred(randval):
-            return randval
-
 def random_bv(width):
     return BitVector[width](random.randint(0, (1 << width) - 1))
-
 
 def random_bit():
     return Bit(random.randint(0, 1))
--- a/fault/random.py
+++ b/fault/random.py
@@ -19,7 +19,7 @@ def random_bv(width):
 def random_bit():
     return Bit(random.randint(0, 1))
 
-class HashableDict(Mapping):
+class FrozenDict(Mapping):
     def __init__(self, d=None):
         if d is None:
             d = dict()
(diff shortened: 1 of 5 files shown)


fault/random.py
```diff
@@ -1,10 +1,246 @@
...
+    def __iter__(self):
+        return iter(self._d)
+
+    def __len__(self):
+        return len(self._d)
+
+    def __hash__(self):
+        if self._hash is None:
+            self._hash = hash(frozenset(self.items()))
+        return self._hash
```
leonardt: It seems like the hash value for an instance is computed and fixed the first time `__hash__` is invoked, is this the intended behavior? What happens if we add items to the dictionary after hash has been called? Or is that a use case that we're not intending to support for this? If so, we should add some docs/guards against this (e.g. calling `__setitem__` after `_hash` has been set could raise an error).

leonardt: Also, `__setitem__` is not defined, but looking at the rest of the code, it seems like this is used to construct a hashable version of a dictionary that isn't modified, so perhaps `FrozenDict` might be a better name?

fault/random.py
```diff
@@ -1,10 +1,246 @@
...
+
+
+    def __call__(
+            self,
+            v_map : tp.Mapping[str, tp.Optional[int]],
+            pred : tp.Callable[..., AbstractBit],
+            N : int,
+            ) -> tp.AbstractSet[tp.Mapping[str, BitVector]]:
+        '''
+        V_map: map of labels to bitvector sizes, size is None indicates a Bit
```
leonardt: Typo, `V_map` -> `v_map` to match lowercase parameter name

fault/random.py
```diff
@@ -1,10 +1,246 @@
...
+        pred: function with signiture (v : AbstractBitVector[w] for v,w in V_map.items()) -> AbstractBit
```
leonardt: Typo `signiture` -> `signature`

fault/random.py
```diff
@@ -1,10 +1,246 @@
...
+                raise TypeError()
+        if solver.check() == z3.sat:
+            model = solver.model()
+            solver.pop()
+            return _model_to_hashdict(v_map, model)
+        else:
+            return None
+
+    def _compute_neighbors(self, v_map, pred, init, seen):
+        conditons = []
```
leonardt: Typo `conditons` -> `conditions`

fault/random.py
```diff
@@ -1,10 +1,246 @@
...
+                elif isinstance(v, AbstractBit):
+                    assignment[k] = random_bit()
+                else:
+                    TypeError()
+
+            assignment = HashableDict(assignment)
+            if assignment not in seen:
+                return assignment
+
+    def _find_closet(self, v_map, pred, seed):
```
leonardt: Typo `_find_closet` -> `_find_closest` (I think?)

cdonovick:
Your institution is correct.  I will add documentation and fix typos.

fault/random.py
cdonovick: __setitem__ is not defined to make the object immutable.  Changing the value of a hashable object will cause strange behavior as they will not be rehashed when they change.  Will change name to FrozenDict

PR #97, opened 2019-05-15
Add basic support for poke/expect X

leonardt:
From what I understand, verilator does not support X (see https://www.veripool.org/projects/verilator/wiki/Manual-verilator#Unknown-states)

This add support for poking/expecting with `fault.UnknownValue` for the `system-verilog` targets (compiled by translating to a Verilog `'X`).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ addons:
   apt:
     packages:
     - verilator
+    - iverilog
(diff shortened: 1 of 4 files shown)


leonardt:
Closing in favor of #98 which has some required changes to the SV target to work with iverilog.

PR #98, opened 2019-05-15
Add basic support for SV file io and poking X values

leonardt:
Fills out basic support for file IO actions in the system verilog backend, based on the `poke-x` branch because I needed some SV target related changes


Kuree:
You use Ubuntu 16 image for `iverilog`. It should have a fairly recent version.

PR #100, opened 2019-05-15
set pipefail when running verilator command

leonardt:
Fixes sub issue in https://github.com/StanfordAHA/lassen/issues/66 where the verilator command was failing silently


leonardt:
Changed base to sv-file-io so i could use iverilog

Kuree:
@leonardt Is it okay to merge now?

leonardt:
@Kuree almost, working through a vcs issue

PR #101, opened 2019-05-15
Fix logic for signed expect/poke

leonardt:
(based on patch-pipefail)

Fixes the logic for compiling signed values to verilator. Basically, verilator always uses the unsigned representation as a C data type, so we need to convert all negative values into their unsigned bits representation. The logic already existed, but was too conservative (it only checked if the port was type as a signed value, but there are cases where expect a signed value on an unsigned port, such as in the case of the signed mode of the PE)

### DIFF
--- a/fault/random.py
+++ b/fault/random.py
@@ -1,8 +1,9 @@
+
 import typing as tp
 import random
 import itertools as it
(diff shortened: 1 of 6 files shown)


Kuree:
Is it okay to merge?

PR #105, opened 2019-05-21
fix input values due to the change for signed values

Kuree:
image

Whenever the high bit for the input is high, i.e. when the value is larger than `0x80`, the current fault will pad 1's to make an negative number. This patch use masks to mask them off.

That's the cause that why `verilator` failed in `GarnetFlow` but `ncsim` works: https://github.com/StanfordAHA/GarnetFlow/commits/master

EDIT:
Travis version is working with this fix: https://travis-ci.com/StanfordAHA/GarnetFlow/builds/112535673

### DIFF
--- a/fault/verilator_target.py
+++ b/fault/verilator_target.py
@@ -178,7 +178,8 @@ class VerilatorTarget(VerilogTarget):
         else:
             value = action.value
             if isinstance(value, actions.FileRead):
-                value = f"*{value.file.name_without_ext}_in"
+                mask = "FF" * value.file.chunk_size
+                value = f"(*{value.file.name_without_ext}_in) & 0x{mask}"
             value = self.process_signed_values(action.port, value)
             result = [f"top->{name} = {value};"]
             # Hack to support verilator's semantics, need to set the register


leonardt:
Interesting, is this because without the mask, C is promoting the byte (char) to a 32 bit type (used by verilator?) which results in sign extension? Then adding the mask results in the promotion to 32 bits without extension because the mask numeric literal is treated as a 32 bit value?

Kuree:
I believe so. `char` in `C/C++` is signed by default. So there might be a sign extension during the implicit conversation. Bitwise `&` should correct this error regardless whether the input is signed or not.

leonardt:
Confirmed via testing!
```c
// promotion.c
#include "printf.h"


int main() {
    char byte = 0xEE;
    printf("%%hhx byte = %hhx\n", byte);
    printf("%%x   byte = %x\n", byte);
    printf("%%x   (int) byte = %x\n", (int) byte);
    printf("%%x   byte & 0xFF = %x\n", byte & 0xFF);
}
```
and
```
/tmp
❯ gcc promotion.c && ./a.out
%hhx byte = ee
%x   byte = ffffffee
%x   (int) byte = ffffffee
%x   byte & 0xFF = ee
```

leonardt:
Another thing we could try is using `unsigned char` instead of `char` for the input type of files. This design enforces that files are read as unsigned bytes. Are there cases where it makes sense to read file data as a signed byte ever?

Kuree:
I need to confirm this with @jeffsetter and other people who do the application development. For DNN the inputs may be signed depends on how they process it.

jeffsetter:
I was templating my testbench to support different types, including the signed and unsigned variants. Images support unsigned and signed types.

PR #109, opened 2019-05-24
Add basic support for expressions

leonardt:
This allows you to expect that an output is a function (expression) of inputs. This adds support for just the And expression, but shows the required extensions to support this in the targets. Will continue to work on filling out the supported operators as well as integrating with while loops (loop until a condition expression).


leonardt:
Closing in favor of https://github.com/leonardt/fault/pull/110

PR #110, opened 2019-05-24
Add basic support for while loop with an expression as the condition (and if statements)

leonardt:


leonardt:
Note this currently doesn't support expressions of arbitrary depth (only supports 1 deep expressions), but that should be an easy extension.
 
Also, there's a lot of duplicated code between the system-verilog and verilator backend, we can likely factor it into `verilog_target` (assuming the operator syntax is identical between the two languages).

PR #111, opened 2019-05-28
only create max_bits once at the top level

THofstee:


THofstee:
My branch pretty regularly merges master so I'm not sure, might just be the
volume of things I'm testing.

PR #115, opened 2019-06-04
Fix file IO with chunk_size > 1

Kuree:
- Using shifts instead of array for larger inputs
- Should support arbitrary chunk size
- Tested with `chunk_size=2` with GarnetFlow

Is there any easy way to create a wide input circuit and test in fault?


leonardt:
Re: Is there any easy way to create a wide input circuit and test in fault?

You can just create a magma circuit with whatever width you'd like, e.g.

```python
import magma as m

class WidePassthroughCircuit(m.Circuit):
    IO = ["I", m.In(m.Bits[128]), "O", m.Out(m.Bits[128])]
    @classmethod
    def definition(io):
        io.O <= io.I
```

PR #117, opened 2019-06-07
Add support for scanf and vars

leonardt:
Adds support for reading in formatted data from files using the `fscanf` interface from C. Relies on a new feature `Var(name, type_)` that allows the declaration of variables. Right now only `BitVector[32]` type_ works for verilator (need to generalize mapping from arbitrary fixed witdth vectors to C code, probably best to leverage Dillon's bit vector library used for the coreir simulator). The system verilog backend supports arbitrary width bit vector types (but not other types) (it generates a `reg [{BV.size -1}:0]`).

Otherwise, it adds a new action `file_scan_format` which accepts a file, a format string, a variable number of variables (basically the C interface).

@THofstee you may be interested in using the Var feature.

Here's the example usage (copied from the test that was checked in)
```python
def test_tester_file_scanf(target, simulator):
    circ = common.TestUInt32Circuit
    tester = fault.Tester(circ)
    tester.zero_inputs()
    file_in = tester.file_open("test_file_in.txt", "r")
    config_addr = tester.Var("config_addr", BitVector[32])
    config_data = tester.Var("config_data", BitVector[32])
    loop = tester.loop(8)
    loop.file_scanf(file_in, "%x %x", config_addr, config_data)
    loop.poke(circ.I, config_addr)
    loop.eval()
    loop.expect(circ.O, config_addr)
    loop.poke(circ.I, config_data)
    loop.eval()
    loop.expect(circ.O, config_data)
    tester.file_close(file_in)
    with tempfile.TemporaryDirectory() as _dir:
        _dir = "build"
        with open(_dir + "/test_file_in.txt", "w") as file:
            file.write(hex(int(BitVector.random(32)))[2:])
        if target == "verilator":
            tester.compile_and_run(target, directory=_dir, flags=["-Wno-fatal"])
        else:
            tester.compile_and_run(target, directory=_dir, simulator=simulator)
```

### DIFF
--- a/tutorial/test.py
+++ b/tutorial/test.py
@@ -1,0 +1,17 @@
+import magma as m
+import fault
+
+
+class Passthrough(m.Circuit):
+    IO = ["I", m.In(m.Bit), "O", m.Out(m.Bit)]
+
+    @classmethod
+    def definition(io):
+        io.O <= io.I
+
+
+passthrough_tester = fault.Tester(Passthrough)
+passthrough_tester.circuit.I = 1
+passthrough_tester.eval()
+passthrough_tester.circuit.O.expect(1)
+passthrough_tester.compile_and_run("verilator")


leonardt:
Going to rebase on https://github.com/leonardt/fault/pull/110 so we I can make `Var` support expressions, also I think @Kuree needs that branch too, so I'm going to pull into that branch and bring that all into master

PR #118, opened 2019-06-07
Add support for scanf and vars

leonardt:
Adds support for reading in formatted data from files using the `fscanf` interface from C. Relies on a new feature `Var(name, type_)` that allows the declaration of variables. Right now only `BitVector[32]` type_ works for verilator (need to generalize mapping from arbitrary fixed witdth vectors to C code, probably best to leverage Dillon's bit vector library used for the coreir simulator). The system verilog backend supports arbitrary width bit vector types (but not other types) (it generates a `reg [{BV.size -1}:0]`).

Otherwise, it adds a new action `file_scan_format` which accepts a file, a format string, a variable number of variables (basically the C interface).

@THofstee you may be interested in using the Var feature.

Here's the example usage (copied from the test that was checked in)
```python
def test_tester_file_scanf(target, simulator):
    circ = common.TestUInt32Circuit
    tester = fault.Tester(circ)
    tester.zero_inputs()
    file_in = tester.file_open("test_file_in.txt", "r")
    config_addr = tester.Var("config_addr", BitVector[32])
    config_data = tester.Var("config_data", BitVector[32])
    loop = tester.loop(8)
    loop.file_scanf(file_in, "%x %x", config_addr, config_data)
    loop.poke(circ.I, config_addr)
    loop.eval()
    loop.expect(circ.O, config_addr)
    loop.poke(circ.I, config_data)
    loop.eval()
    loop.expect(circ.O, config_data)
    tester.file_close(file_in)
    with tempfile.TemporaryDirectory() as _dir:
        _dir = "build"
        with open(_dir + "/test_file_in.txt", "w") as file:
            file.write(hex(int(BitVector.random(32)))[2:])
        if target == "verilator":
            tester.compile_and_run(target, directory=_dir, flags=["-Wno-fatal"])
        else:
            tester.compile_and_run(target, directory=_dir, simulator=simulator)
```


leonardt:
There were some collateral changes to README/tutorial/examples on master, I managed to get everything in sync via rebase, but for some reason github still shows them in the files changed, sorry about that to reviewers but can't seem to find a reasonable way to get github to not show them.

leonardt:
@Kuree FYI this also changes how chunk size is handled, there seemed to be inconsistency with byte ordering. I'm not sure if you're using a chunk_size >= 1, but if you are, there might be a bug introduced by these changes if you were relying on the old byte ordering. See these changes for more info https://github.com/leonardt/fault/pull/118/commits/3a5496a3915908d805c50dd3b93334c409eed1f4#diff-998b6b3cdd76a92ad86257071e611c12R191

leonardt:
Re: chunk size issue this came up when changing the verilator implementation to use the C file i/o (before it was using the c++ fstream), so this may also affect any verilator test bench flows.

Kuree:
I am using `chunk_size > 1` for the floating point tests. Since the floating point is not working at the moment, we can fix it once the problem comes up.

leonardt:
Okay, let me know if/when you run into problems. We can also change the byte ordering back to what it was if that makes more sense to you.

leonardt:
(the main problem was that it was inconsistent, which ordering we choose matters less to me)

Kuree:
I use little endian throughout the simulation.

PR #119, opened 2019-06-12
Add support for both endian modes

leonardt:
Reverts default file IO behavior to be little endian (the semantics got mixed up because Pythons `file.write` method is weird and caused some semantic mismatch in the expected behavior and actual behavior) and adds a parameter to file open to select endianness.

### DIFF
--- a/fault/system_verilog_target.py
+++ b/fault/system_verilog_target.py
@@ -190,7 +190,6 @@ if (!{name}_file) $error("Could not open file {action.file.name}: %0d", {name}_f
             loop_expr = f"__i = {action.file.chunk_size - 1}; __i >= 0; __i--"
         else:
             loop_expr = f"__i = 0; __i < {action.file.chunk_size}; __i++"
-        # notice that this is little endian
         code = f"""\
 {action.file.name_without_ext}_in = 0;
 for ({loop_expr}) begin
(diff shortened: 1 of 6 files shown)


fault/system_verilog_target.py
```diff
@@ -186,10 +186,14 @@ def make_file_read(self, i, action):
         decl = f"integer __i;"
         if decl not in self.declarations:
             self.declarations.append(decl)
+        if action.file.endianness == "big":
+            loop_expr = f"__i = {action.file.chunk_size - 1}; __i >= 0; __i--"
+        else:
+            loop_expr = f"__i = 0; __i < {action.file.chunk_size}; __i++"
         # notice that this is little endian
```
THofstee: `# notice that this is little endian` comment seems outdated.

fault/system_verilog_target.py
```diff
@@ -201,9 +205,13 @@ def make_file_write(self, i, action):
         decl = f"integer __i;"
         if decl not in self.declarations:
             self.declarations.append(decl)
+        if action.file.endianness == "big":
+            loop_expr = f"__i = {action.file.chunk_size - 1}; __i >= 0; __i--"
+        else:
+            loop_expr = f"__i = 0; __i < {action.file.chunk_size}; __i++"
         # this is little endian as well
```
THofstee: `# this is little endian as well` also seems outdated.

PR #139, opened 2019-07-31
Working with external verilog files

sgherbst:
These commits span updates to the ncsim, vcs, and iverilog simulators in SystemVerilogTarget to make it easier to work with external files:
1. There is no file added to the source file list for the DUT itself when **ext_model_file=True**.  This, combined with the existing **ext_libs** option, gives the user an option to point to external verilog files rather than copying them over to the build directory outside of fault.  Perhaps in the future, when **skip_compile=True**, **ext_model_file** should default to **True** as well.  But I didn't want to break existing programs that depend on the current behavior.
2. The generation of commands for ncsim, vcs, and iverilog are moved into separate methods to make it easier to make simulator-specific updates in the future.
3. The files specified by the **ext_libs** argument are included as libraries ("-v" or "-y", depending on the simulator) rather than directly as source files.  This has the benefit of making their order unimportant.  The behavior of the **include_verilog_libraries** argument is unmodified, although it's worth noting that it really is including files as verilog sources, not libraries.  As an implementation note -- iverilog only supports libdirs, not lib files, so a set of unique directories is extracted from the ext_lib argument.
4. The use of subprocess to call the three simulators is pulled into the method subprocess_run to avoid code duplication for shell escaping, subprocess run options, and logging of outputs.  As a side note, it would be good to investigate why shell=True is needed -- test_power_domains fails without it on Travis but passes on my setup.  For now I left shell=True option in.
5. A new test, test_ext_vlog.py, exercises the above features.

### DIFF
--- a/fault/system_verilog_target.py
+++ b/fault/system_verilog_target.py
@@ -35,7 +35,8 @@ class SystemVerilogTarget(VerilogTarget):
                  skip_compile=False, magma_output="coreir-verilog",
                  magma_opts={}, include_verilog_libraries=[], simulator=None,
                  timescale="1ns/1ns", clock_step_delay=5, num_cycles=10000,
-                 dump_vcd=True, no_warning=False, sim_env=None):
+                 dump_vcd=True, no_warning=False, sim_env=None,
+                 ext_model_file=False):
         """
         circuit: a magma circuit
 
@@ -62,6 +63,11 @@ class SystemVerilogTarget(VerilogTarget):
         sim_env: Environment variable definitions to use when running the
                  simulator.  If not provided, the value from os.environ will
                  be used.
+
+        ext_model_file: If True, don't include the assumed model name in the
+                        list of Verilog sources.  The assumption is that the
+                        user has already taken care of this via
+                        include_verilog_libraries.
         """
         super().__init__(circuit, circuit_name, directory, skip_compile,
                          include_verilog_libraries, magma_output, magma_opts)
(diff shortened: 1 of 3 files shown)


sgherbst:
Not sure why GarnetFlow is failing -- does anyone know what these errors mean?  It doesn't look like the check is getting to the point that it can run fault tests.
```shell
error: patch failed: CMakeLists.txt:20
error: CMakeLists.txt: patch does not apply
error: patch failed: src/passes/analysis/verilog.cpp:2
error: src/passes/analysis/verilog.cpp: patch does not apply
./install.sh: line 13: cd: coreir/build: No such file or directory
python: can't open file 'python_repo.py': [Errno 2] No such file or directory
```

leonardt:
CoreIR just merged a major update to the verilog backend/toolchain, it looks like this broke a downstream tool flow, we will investigate, but this is not your fault.

sgherbst:
Just re-ran the tests after updates to GarnetFlow and it looks like everything is passing now.

leonardt:
`shell=True` is used to make it easier to pass through shell commands. Typically, you need to pass the commands as a list of strings rather than a single command string. There are various reason fo this being the standard interface (programmatically building commands, preventing shell injection) but usually `shell=True` is simpler (re: quicker) to use. It's likely that we can't remove shell=True because there's some part of the command construction process that isn't properly splitting the command line arguments into separate strings.

PR #142, opened 2019-08-01
Remove shell=True, fix test to skip if irun not available

leonardt:


leonardt:
The issue was that the test was not being skipped if `irun` was not available, but turns out the error code wasn't propagating with `shell=True` which is strange

leonardt:
Ah, I see, the test expects an `AssertionError` because it's designed to fail. However, when changing from `shell=True`, we actually get
```
FileNotFoundError: [Errno 2] No such file or directory: 'irun': 'irun'
```
so that wasn't being caught by the `except` statement.

sgherbst:
Looks great, thanks!

PR #143, opened 2019-08-04
Preliminary Verilog-AMS support

sgherbst:
This pull request adds support for mixed-signal simulation via VerilogAMSTarget, which inherits from SystemVerilogTarget.  The following features are supported:
1. Generate a VerilogAMS wrapper for a SPICE/Spectre netlist, which itself is a magma circuit that can be provided as the argument to a Tester instantiation.  This is done using the VAMSWrap command.
2. AnalogIn and AnalogOut are added as port types that may be used in a magma DeclareCircuit command.  This probably should be moved to magma at some point.
3. The new VerilogAMSTarget allows the user to specify locations of SPICE/Spectre files (via **model_paths**) and the voltage+resistance used for D2A and A2D conversions (via **vsup** and **rout**).
4. VerilogAMSTarget automatically generates the AMS control file needed for mixed-signal simulation as part of the run() function, similar to how TCL files are generated in the run() command of SystemVerilogTarget.
5. tests/test_vams_sim.py illustrates these features by instantiating a SPICE-level CMOS inverter (tests/spice/myinv.sp) and then checking its behavior in response to a randomized binary stimulus.
6. tests/test_vams_wrap.py is a more low-level test of the generation of a Verilog-AMS wrapper for a mixed-signal circuit.

### DIFF
--- a/fault/__init__.py
+++ b/fault/__init__.py
@@ -3,6 +3,7 @@ from .power_tester import PowerTester
 from .value import Value, AnyValue, UnknownValue
 import fault.random
 from .symbolic_tester import SymbolicTester
+from .verilogams import VAMSWrap, AnalogIn, AnalogOut
 
 
 class WrappedVerilogInternalPort:
--- a/fault/__init__.py
+++ b/fault/__init__.py
@@ -3,7 +3,8 @@ from .power_tester import PowerTester
 from .value import Value, AnyValue, UnknownValue
 import fault.random
 from .symbolic_tester import SymbolicTester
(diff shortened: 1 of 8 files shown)


sgherbst:
Looks like most tests are failing in the build -- perhaps there were changes to coreir/magma/mantle/etc?  Here are some of the errors:
* AttributeError: /usr/lib/libcoreir-c.so: undefined symbol: COREBitInOut
* The command "pip install mantle" failed and exited with 1 during . (seems that coreir build fails with "fatal error: variant: No such file or directory")

sgherbst:

PR #144, opened 2019-08-05
Working with external testbenches

sgherbst:
This PR allows a user to run an existing (presumably hand-written) SystemVerilog testbench using **fault**.  In this case, **fault** is essentially being using for its capability to abstract simulator commands over ncsim, vcs, and iverilog.  However, this is still a useful capability in its own right, and might be a stepping stone for the user to get started creating their own testbenches using the full power of **fault**.

The following features are included:
1.  Added an argument "ext_test_bench" to SystemVerilogTarget.  If True, no testbench file will be written for fault and the simulator won't be instructed to look for it.
2. Added an argument "inc_dirs" to SystemVerilogTarget.  This is the search path used by the simulator for "`include" statements.
3.  Added an argument "top_module" to SystemVerilogTarget.  Generally this can be left at its default value (None), and it will be filled automatically if needed.  This addition was needed because the ncsim command used to always include "-top {circuit.name}_tb", which isn't necessarily correct in the case of an external test bench.
4. Added an argument "ext_srcs" to SystemVerilogTarget.  It is an alias for the longer argument name "include_verilog_libraries".  A conservative approach is taken to avoid confusion: it is deemed illegal to specify both.
5. Updated the default-picking scheme for "skip_compile", "ext_model_file", and "ext_test_bench": if ext_model_file isn't specified, it defaults to "ext_test_bench".  Similarly, if "skip_compile" isn't specified, it defaults to "ext_model_file".  This keeps backward compatibility while making it easier to work with external models and testbenches.
6. Now using "None" rather than "[]" and "{}" in kwarg defaults in SystemVerilogTarget.__init__ and SystemVerilogTarget.run (the correct defaults are filled in the body of the function itself)
7. Added **test_ext_tb** and **test_inc_dir**.
8. Refactored **test_def_vlog**, **test_ext_vlog**, and **test_while_loop** to take advantage of new features.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ __pycache__
 build
 dist
 fault.egg-info
+vnc_logs
--- a/fault/system_verilog_target.py
+++ b/fault/system_verilog_target.py
@@ -632,7 +632,7 @@ end;
         cmd += ['-LDFLAGS']
         cmd += ['-Wl,--no-as-needed']
         if self.dump_vcd:
-            cmd += ['-debug_access+all']
+            cmd += ['+vcs+vcdpluson', '-debug_pp']
 
         # return arg list and binary file location
         return cmd, './simv'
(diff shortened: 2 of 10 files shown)


sgherbst:
The tests seem to be failing for a similar reason as in pull request #143.  Doesn't seem to be related to the content of the PR because the build is crashing before getting to pytest.

sgherbst:

PR #146, opened 2019-08-07
Features to support transistor-level verification of SRAM-like circuits

sgherbst:
Depends on pull requests #143 and #144.

This pull request adds several features useful in the simulation of SRAM cells (both Verilog models and SPICE implementations):
1. A new value called "HiZ" is added, which is interpreted as the "'Z" value in SystemVerilog simulation.
2. The "expect" command now has an optional "strict" argument that causes the "!==" symbol to be emitted instead of "!=".  Without it, X's slip through undetected, so I would recommend that we set strict=True by default.  However, I have currently left strict=False by default to preserve backwards compatibility until we can investigate the ramifications of this change.
3. The magma.InOut type is now supported in SystemVerilogTarget.  When poking the port, values are written to a register which is in turn assigned to a wire that is connected to the DUT.  When expecting values from the port, the values are read off the wire directly.
4. An option "use_input_wires" is added to SystemVerilogTarget, mainly for use by its subclass, VerilogAMSTarget.  When enabled, all DUT inputs are connected to wires that are in turn assigned to registers.  This makes SPICE and VerilogAMS blocks play more nicely with the testbench, but isn't needed for purely digital simulation.  Hence the default is False for SystemVerilogTarget and True for VerilogAMSTarget.
5. Added an option "use_spice" to VerilogAMSTarget that allows the users to list out the modules that should be simulated at spice level explicitly.
6. Added tests test_bidir.py, test_hi_z.py, and test_vams_sram.py to check the above features and updated other tests as necessary.


sgherbst:

PR #148, opened 2019-08-09
Mixed-signal simulation, external testbenches, SRAM bits, and real numbers

sgherbst:
## Summary 

This pull request adds several new features:
1. Support for mixed-signal simulation via VerilogAMSTarget, which inherits from SystemVerilogTarget.  The corresponding target is called 'verilog-ams'.
2. Support running existing (presumably hand-written) SystemVerilog testbenches using **fault**.  In this case, **fault** is essentially being using for its capability to abstract simulator commands over ncsim, vcs, and iverilog.  However, this is still a useful capability in its own right, and might be a stepping stone for the user to get started creating their own testbenches using the full power of **fault**.
3.  Support for simulating SRAM bits (both Verilog models and SPICE implementations) is added, along with a subclass of Tester called SRAMTester that implements a basic test of SRAM functionality.  One of the main features added was support for Hi-Z signals (this is important for SRAM because bit lines are bi-directional)
4. Support for real numbers in SystemVerilogTarget and VerilogAMSTarget, mainly by expanding the functionality of the "expect" statement to work with different kinds of real-number ranges.

Details of these high-level features are included in the sections below.

## VerilogAMS support

1. Generate a VerilogAMS wrapper for a SPICE/Spectre netlist, which itself is a magma circuit that can be provided as the argument to a Tester instantiation.  This is done using the VAMSWrap command.
2. AnalogIn and AnalogOut are added as port types that may be used in a magma DeclareCircuit command.  This probably should be moved to magma at some point.
3. The new VerilogAMSTarget allows the user to specify locations of SPICE/Spectre files (via **model_paths**) and the voltage+resistance used for D2A and A2D conversions (via **vsup** and **rout**).
4. VerilogAMSTarget automatically generates the AMS control file needed for mixed-signal simulation as part of the run() function, similar to how TCL files are generated in the run() command of SystemVerilogTarget.
5. tests/test_vams_sim.py illustrates these features by instantiating a SPICE-level CMOS inverter (tests/spice/myinv.sp) and then checking its behavior in response to a randomized binary stimulus.
6. tests/test_vams_wrap.py is a more low-level test of the generation of a Verilog-AMS wrapper for a mixed-signal circuit.

## Support for external testbenches

1.  Added an argument "ext_test_bench" to SystemVerilogTarget.  If True, no testbench file will be written for fault and the simulator won't be instructed to look for it.
2. Added an argument "inc_dirs" to SystemVerilogTarget.  This is the search path used by the simulator for "`include" statements.
3.  Added an argument "top_module" to SystemVerilogTarget.  Generally this can be left at its default value (None), and it will be filled automatically if needed.  This addition was needed because the ncsim command used to always include "-top {circuit.name}_tb", which isn't necessarily correct in the case of an external test bench.
4. Added an argument "ext_srcs" to SystemVerilogTarget.  It is an alias for the longer argument name "include_verilog_libraries".  A conservative approach is taken to avoid confusion: it is deemed illegal to specify both.
5. Updated the default-picking scheme for "skip_compile", "ext_model_file", and "ext_test_bench": if ext_model_file isn't specified, it defaults to "ext_test_bench".  Similarly, if "skip_compile" isn't specified, it defaults to "ext_model_file".  This keeps backward compatibility while making it easier to work with external models and testbenches.
6. Now using "None" rather than "[]" and "{}" in kwarg defaults in SystemVerilogTarget.__init__ and SystemVerilogTarget.run (the correct defaults are filled in the body of the function itself)
7. Added **test_ext_tb** and **test_inc_dir**.
8. Refactored **test_def_vlog**, **test_ext_vlog**, and **test_while_loop** to take advantage of new features.

## Support for SRAM cells

1. A new value called "HiZ" is added, which is interpreted as the "'Z" value in SystemVerilog simulation.
2. The "expect" command now has an optional "strict" argument that causes the "!==" symbol to be emitted instead of "!=".  Without it, X's slip through undetected, so I would recommend that we set strict=True by default.  However, I have currently left strict=False by default to preserve backwards compatibility until we can investigate the ramifications of this change.
3. The magma.InOut type is now supported in SystemVerilogTarget.  When poking the port, values are written to a register which is in turn assigned to a wire that is connected to the DUT.  When expecting values from the port, the values are read off the wire directly.
4. An option "use_input_wires" is added to SystemVerilogTarget, mainly for use by its subclass, VerilogAMSTarget.  When enabled, all DUT inputs are connected to wires that are in turn assigned to registers.  This makes SPICE and VerilogAMS blocks play more nicely with the testbench, but isn't needed for purely digital simulation.  Hence the default is False for SystemVerilogTarget and True for VerilogAMSTarget.
5. Added an option "use_spice" to VerilogAMSTarget that allows the users to list out the modules that should be simulated at spice level explicitly.
6. Added a subclass of Tester called SRAMTester that implements a test of SRAM bit functionality.  It is written in a way such that tester can be applied to both Verilog model and SPICE implementations, and has a little bit of flexibility in terms of the presence of supply pins and their names.
7. Added tests test_bidir.py, test_hi_z.py, and test_vams_sram.py to check the above features and updated other tests as necessary.

## Support for real numbers

1. Added optional arguments "above", "below", "rel_tol", and "abs_tol" to the **expect** action for use in real number comparisons, and updated SystemVerilogTarget to implement various kinds of assertions depending on these arguments.
2. Added I/O types RealIn, RealOut, and RealInOut to describe real-number ports, and updated SystemVerilogTarget to generate the appropriate port definitions for these signals.  (ElectIn, ElectOut, and ElectInOut are also added, but are only implemented for the Verilog-AMS wrapper -- there is no equivalent for SystemVerilog)
3. Added test_real_val.py and test_inv_tf.py to check these features.  **test_inv_tf** is a lower-level implementation of an inverter functionality check, which checks that the voltages measured at the output of the inverter match what is expected given the voltages at the input.


fault/__init__.py
```diff
@@ -1,8 +1,12 @@
+from .real_type import RealIn, RealOut
+from .elect_type import ElectIn, ElectOut
 from .tester import Tester
 from .power_tester import PowerTester
-from .value import Value, AnyValue, UnknownValue
+from .value import Value, AnyValue, UnknownValue, HiZ
 import fault.random
 from .symbolic_tester import SymbolicTester
+from .verilogams import VAMSWrap
+from .tester_samples import *
```
leonardt: Can we change this from an `import *` to explicitly importing what you want? In practice, I prefer the style of explicit imports as it's easier to track where things are coming from (avoids bugs where something is accidentally imported when it wasn't meant to be, possibly clobbering an existing name)

fault/elect_type.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+def MakeElect(**kwargs):
+    return ElectKind('Elect', (ElectType,), kwargs)
+
+
+Elect = MakeElect()
+ElectIn = MakeElect(direction=INPUT)
+ElectOut = MakeElect(direction=OUTPUT)
+ElectInOut = MakeElect(direction=INOUT)
```
leonardt: Seems like we could provide this in magma, but we can leave it in fault for now to experiment with it and merge it upstream when it's more mature (also we're planning a rewrite of the magma type system so it may be better to wait until that's complete, I will include this in the new type hierarchy).

fault/real_type.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+def MakeReal(**kwargs):
+    return RealKind('Real', (RealType,), kwargs)
+
+
+Real = MakeReal()
+RealIn = MakeReal(direction=INPUT)
+RealOut = MakeReal(direction=OUTPUT)
+RealInOut = MakeReal(direction=INOUT)
```
leonardt: Same comment as above on ElecType.

leonardt:
Actually it seems like those tests are being run under ncsim, so they are being checked in the buildkite flow.

fault/__init__.py
sgherbst: Sorry about that -- will fix it in this week's PR.

fault/elect_type.py
sgherbst: Great, thanks :-)

PR #150, opened 2019-08-19
SpiceTarget, user config, and increasing modularity

sgherbst:
## Introduction
The biggest feature added in this PR is a new target to support spice simulation.  A user config file to specify for environment settings is also added, enabling shorting and more reusable tests.  Finally, refactoring in several areas (hopefully) improves modularity and makes future extensions easier.  Details on all of these changes are included below.

## SpiceTarget
1. Added a SpiceTarget, supporting ngspice (open source), hspice (commercial), and spectre (commercial) simulators.  Only the Poke, Expect, and Delay actions are currently implemented.  Rather than using in-simulation assertions like the SystemVerilogTarget, Expect actions are implemented by post-processing simulator results (using the nutascii parser for ngpisce and spectre and PSF parser for HSPICE).  Poke actions are implemented by constructing PWL waveform tables.  For InOut ports, it is legal to drive them to fault.HiZ, as is allowed in SystemVerilogTarget and VerilogAMSTarget.  In the spice-simulation case, InOut ports are implemented with a switch model in between a PWL voltage source and the DUT pin of interest.
2. Added **Delay** action, which is invoked via Tester.delay(<time in seconds>).  Currently implemented only for the SpiceTarget, but the implementation should be straightforward for SystemVerilogTarget (emit "#(delay*1s)").
3. Added netlister utility for converting Cadence Virtuoso schematic-based designs into spice netlist.  The netlister is invoked from fault.netlister.si_netlist; the user can specify things like the library name, cell name, and cds.lib location.  There is currently no regression test for this feature owing to the challenge of creating a test design that can be released as open source.  It might be worth exploring whether [FreePDK](https://www.eda.ncsu.edu/wiki/FreePDK) could be used in constructing open-source test designs in the future.
4. Relevant tests are test_spice_bus, test_result_parsing, test_pwl_gen.  test_inv_tf and test_logic are also updated to check the spice target.  Note that ngspice will need to be installed in the Travis script to check this -- "apt-get install ngspice" and "brew install ngspice" should work.  For the BuildKite server, hspice and/or spectre can likely be made available via a "module load" command.

## User Config
1. Added an optional fault config file which can live in the working directory at "fault.yml" or in the home directory at "~/.faultrc".  The syntax is YAML, and currently supported options are things like removing conda from the env variables sent to the simulator, and adding certain define variables to the same set of env variables.  These changes do not modify the user's shell environment.

## Increasing Modularity
1. Refactored common parts of Spice and VerilogAMS code generation into a more generic class called CodeGenerator.  This could be applied to SystemVerilogTarget as well, but I haven't done this yet.
2. Refactored subprocess run-and-display code into a common location used by both SystemVerilogTarget and SpiceTarget.
3. Refactored common functions used by pytest tests to further consolidate code that appeared in multiple places and to make PyCharm happier.

## Miscellaneous
1. Added the capability to instantiate a SystemVerilog DUT with parameters (relevant if the DUT is hand-coded).  Just pass a dict mapping parameter names to values via the "parameters" argument when calling "compile_and_run".
2. Extended the Print action in SystemVerilogTarget to support printing literals that are already defined in Python.  This is a convenience to avoid having to mix Python syntax and SystemVerilog syntax in the format string.


fault/tester_samples.py
```diff
@@ -1,46 +1,131 @@
...
+
+    def expect(self, name, *args, **kwargs):
+        super().expect(getattr(self._circuit, name), *args, **kwargs)
+
+    def define_init(self):
+        self.poke_optional(self.supply1, 1)
+        self.poke_optional(self.supply0, 0)
+
+    def define_trial(self):
+        pass
```
leonardt: You could make this an abstractmethod (need to use ABCMeta too, see https://docs.python.org/3/library/abc.html#abc.abstractmethod) to make sure that derived classes implement it.

fault/tester_samples.py
```diff
@@ -1,46 +1,131 @@
...
+    def __init__(self, circuit, *args, inputs=None, out='out', **kwargs):
+        # save settings
+        self.inputs = inputs if inputs is not None else []
+        self.out = out
+
+        # call super constructor
+        super().__init__(circuit, *args, **kwargs)
+
+    def model(self, *args):
+        raise NotImplementedError
```
leonardt: Same as above (abc pattern)

fault/tester_samples.py
sgherbst: Thanks, will include this in the next PR

PR #151, opened 2019-08-24
Real-time subprocess output; initial conditions and selective probing for Spice.

sgherbst:
## Introduction

The biggest change from this PR is more unified handling of subprocess calls for SystemVerilog, Verilator, VerilogAMS, and spice targets, through a "deluxe" version of subprocess.run that supports real-time output.  Other updates include exposing some more spice-level simulation capabilities and adding more examples.  Details included below.

## fault.subprocess_run.subprocess_run
1. Added capability to print out STDOUT lines as they come in.  Lines can be printed using logging or regular print(), depending on the optional **disp_type** argument.  This is useful for long simulations in which realtime output is helpful.  (STDERR is printed out in its entirety after the process run completes.)
2. To make the output easy to read, prepended logging info is temporarily disabled while running subprocess_run (otherwise every line would be prepended with "INFO <line info>".  In addition terminal coloring is now used to indicate the start of a new subprocess run and STDOUT vs. STDERR output sections.  The user can disable all output by setting **disp_type** to **None** or via the **logging** level.
3. Error checking is embedded in **subprocess_run** through the optional **err_str** argument. 
 When this argument is provided, each line of STDOUT and STDERR will be search to see if it contains the provided error string, raising an AssertionError if present.  In the future, it would be good to explore allowing regex, but for the existing tests a single **err_str** seems to be sufficient.
4. When **check_ret_code** is **True** (default), **subprocess_run** will raise an **AssertionError** if the subprocess exits with a non-zero code.
5. Updated VerilatorTarget to make use of new subprocess_run functions, making the handling of subprocess runs more standardized.

## SpiceTarget
1. Add selective probing to SpiceTarget: this cuts down the amount to data saved during spice simulation to just the minimum required needed to implement Print and Expect
2. Add capability to set initial conditions for spice and verilog-ams targets (tested by tests/test_init_cond.py).  The main difficulty is is specifying the path to nodes that should be initialized in a target-agnostic manner.  For this I have added Tester.internal, which is sort of like an os.path.join for producing SelectPaths.  In the future I hope to better integrate this with things like WrappedVerilogPort.
3. Fix convergence issues in switch model used for target=spice, simulator=ngspice

## Other
1. Add minimal example for spice simulation: examples/myinv.py
2. Add example for determine SRAM noise margin: examples/sram_snm.py
3. Add optional defaults to Poke and Expect at the tester level: Expects can now be made "strict" by default and Poke default delays can be set to any value (most importantly, "0") through optional arguments to the Tester constructor.

### DIFF
--- a/fault/spice_target.py
+++ b/fault/spice_target.py
@@ -13,11 +13,18 @@ from fault.pwl import pwc_to_pwl
 from fault.actions import Poke, Expect, Delay
 
 
+# define a custom error for A2D conversion to make it easier
+# to catch this specific issue (for example, in adapting
+# tests based on the previous test result)
+class A2DError(Exception):
+    pass
+
+
 class SpiceTarget(Target):
     def __init__(self, circuit, directory="build/", simulator='ngspice',
                  vsup=1.0, rout=1, model_paths=None, sim_env=None,
                  t_step=None, clock_step_delay=5, t_tr=0.2e-9, vil_rel=0.4,
-                 vih_rel=0.6, rz=1e9, conn_order='alpha', bus_delim='<',
+                 vih_rel=0.6, rz=1e9, conn_order='alpha', bus_delim='<>',
                  bus_order='descend', flags=None):
         """
         circuit: a magma circuit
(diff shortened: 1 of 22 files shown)


fault/spice_target.py
```diff
@@ -10,15 +10,32 @@
...
+    pass
+
+
+class CompiledSpiceActions:
+    def __init__(self, pwls, checks, prints, stop_time, saves):
+        self.pwls = pwls
+        self.checks = checks
+        self.prints = prints
+        self.stop_time = stop_time
+        self.saves = saves
```
leonardt: Python 3.7's dataclass is a new feature that makes writing classes like this simpler (assuming your intended use pattern fits into the model), check out
* https://realpython.com/python-data-classes/
* https://docs.python.org/3/library/dataclasses.html

Issue #154, opened 2019-08-29
Support Expect With Tuple Value


David-Durst:
I'd like the following psuedocode to work:
```
# let tester.circuit.O0's type be Tuple({"k": Array[5, BitIn], "v": Array[2, BitIn]})
input = {"k" : 4, "v" : 2}
tester.circuit.O0.expect(input1.k)
```

David-Durst:
Addressed by https://github.com/leonardt/fault/pull/156.

PR #156, opened 2019-08-30
Add support for poke/expect dictionary for tuples

leonardt:
Addresses https://github.com/leonardt/fault/issues/154

See https://github.com/leonardt/fault/compare/tuple-actions?expand=1#diff-de403c371b3cd6c4725f37fb3a2fe531R548-R592 for example usage

### DIFF
--- a/fault/tester.py
+++ b/fault/tester.py
@@ -111,10 +111,20 @@ class Tester:
         if delay is None:
             delay = self.poke_delay_default
 
+        def recurse(port):
+            if isinstance(value, dict):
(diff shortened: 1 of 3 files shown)


leonardt:
It supports poking using unnamed tuples `(a, b, c, ...)` and does it based on ordering, and a dictionary with key/value mappings for named tuples.

leonardt:
Ideally we'd throw errors when there's a mismatch between the value and the port type (e.g. a dictionary for an unnamed tuple), but this should give you the desired feature.

Issue #160, opened 2019-09-20
Expect Array Of Tuples Fails


David-Durst:
https://github.com/David-Durst/aetherling/blob/14ef425256d53f05f4f2edc853cdcd2783ad1099/aetherling/HaskellPrelude.py#L68 fails with 
```
Traceback (most recent call last):
  File "/Users/durst/dev/W17-8/aetherling/aetherling/HaskellPrelude.py", line 68, in <module>
    tester.circuit.O.expect(fault_output[f_clk])
  File "/Users/durst/dev/W17-8/fault/fault/wrapper.py", line 60, in expect
    select_path.tester.expect(select_path, value)
  File "/Users/durst/dev/W17-8/fault/fault/tester.py", line 173, in expect
    value = make_value(port, value)
  File "/Users/durst/dev/W17-8/fault/fault/value_utils.py", line 22, in make_value
    return make_array(switch.T, switch.N, value)
  File "/Users/durst/dev/W17-8/fault/fault/value_utils.py", line 52, in make_array
    raise NotImplementedError(T, N, value)
NotImplementedError: (Tuple(0: Array[8, In(Bit)], 1: Array[8, In(Bit)]), 4, [(0, 0), (1, 1), (2, 2), (3, 3)])
```

The issue appears to be with https://github.com/leonardt/fault/blob/7aa6353cbbb42c9d5f26fa2f8f848714f6a5ce0b/fault/value_utils.py#L24.

When I try manually creating the expects for each element of the tuple, I get:
```
Running command: make -C obj_dir -j -f VModule_2.mk VModule_2
<STDOUT>
clang++  -I.  -MMD -I/usr/local/Cellar/verilator/4.014/share/verilator/include -I/usr/local/Cellar/verilator/4.014/share/verilator/include/vltstd -DVL_PRINTF=printf -DVM_COVERAGE=0 -DVM_SC=0 -DVM_TRACE=0 -faligned-new -fbracket-depth=4096 -Qunused-arguments -Wno-parentheses-equality -Wno-sign-compare -Wno-uninitialized -Wno-unused-parameter -Wno-unused-variable -Wno-shadow       -c -o Module_2_driver.o ../Module_2_driver.cpp
[...]

PR #162, opened 2019-09-24
Improve handling of nested types

leonardt:
Fixes https://github.com/leonardt/fault/issues/160 by improving the tester logic to recurse when handling general arrays (this logic was added for tuples, but not arrays). Now the tester logic should properly recurse for both array and tuple types.


David-Durst:
lgtm

PR #180, opened 2019-12-06
Remove extraneous timesteps in action generation

leonardt:
Before, fault was inserting #5 delays for every poke action, as a result the timing of the waveforms was off as it shows each poke action occurring during a different timestep.  I'm not sure why this was included, but removing it makes the timing more appropriate (delays only occur with clock events) and the tests pass.

There is still a `delay` parameter for poke actions that can be supplied explicitly by the user, this just removes the default delay for all pokes.

Also adds a small change to support vcd generation with iverilog

### DIFF
--- a/fault/spice_target.py
+++ b/fault/spice_target.py
@@ -10,7 +10,7 @@ from fault.nutascii_parse import nutascii_parse
 from fault.psf_parse import psf_parse
 from fault.subprocess_run import subprocess_run
 from fault.pwl import pwc_to_pwl
-from fault.actions import Poke, Expect, Delay, Print
+from fault.actions import Poke, Expect, Delay, Print, Eval
 from fault.select_path import SelectPath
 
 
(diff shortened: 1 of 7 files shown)


leonardt:
An issue arises when the user specifies a poke, followed by an expect.  In this case, they happen during the same timestep, which could lead to an expected ordering.  Currently I'm proposing that we use the notion of `eval` (which is used in the verilator side of the things to trigger a computation before expect) to insert a `#0` which should make the expect events happen after the poke events.

leonardt:
The issue is that without some delay between a poke and expect event, the system verilog test benches would sample the wrong value for the expect (before the poke).  To remedy this, calling `eval` inserts a #1, which causes the subsequent expect actions to happen after the poke events have been processed.  This mimics the verilator behavior.  For clocked test benches, there's an implicit #{clock_step_delay} for `step` actions which also handle this issue.

PR #182, opened 2019-12-09
Add warning for no initialized clock

leonardt:
Catches issue when clock is initialized to X for system verilog test benches.  In the future, we will turn this into an error, but for now we just issue a warning so we don't break existing code.

### DIFF
--- a/fault/tester.py
+++ b/fault/tester.py
@@ -263,7 +263,8 @@ class Tester:
         """
         if not self.clock_initialized:
             warnings.warn("Clock has not been initialized, the initial value "
-                          "will be X for system verilog targets")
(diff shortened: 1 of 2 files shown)


rsetaluri:
Should the logic be that we check `clock_initialized` before `.step()` is called, not just before `compile()` ?

leonardt:
That issue with that would be that it would trigger a bunch of warnings if the test bench calls step many times, I think it's probably better to just issue one warning in the case that this happens.

leonardt:
Although we could add the logic to step to only warn once

leonardt:
Actually, that makes sense to catch the case that they user pokes the clock after stepping.

PR #184, opened 2019-12-10
Add .buildkite/pipeline.yml

sgherbst:
This PR adds a YAML file containing the BuildKite test configuration.  Our BuildKite infrastructure has not be updated to use this file yet, but by merging this PR we should be able to make that change as seamlessly as possible.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,0 +1,35 @@
+steps:
+  - command: |
+      # set up environment
+      export PATH=/usr/local/miniconda/bin:/usr/local/bin:$$PATH
+      echo $$PATH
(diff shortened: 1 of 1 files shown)


.buildkite/pipeline.yml
```diff
@@ -0,0 +1,35 @@
+steps:
+  - command: |
+      # set up environment
+      export PATH=/usr/local/miniconda/bin:/usr/local/bin:$$PATH
+      echo $$PATH
+      export LD_LIBRARY_PATH=$$LD_LIBRARY_PATH:/usr/local/lib
+      echo $$LD_LIBRARY_PATH
```
leonardt: IIRC, the double $$ were required to escape the $ symbol in the scripting interface.  Just in case this pops up in debugging (not sure if it's the same syntax for the yaml interface)

PR #185, opened 2019-12-10
Add support for Vivado as a simulator

sgherbst:
This PR adds a new option for running SystemVerilog simulations using Vivado.  This option can be used by setting simulator='vivado' when invoking **compile_and_run**.

Vivado support is tested as part of the regression flow using the existing fpga-verif-* BuildKite agents.

Details:
1. Since Vivado can produce one of several error strings, the **subprocess_run** command was updated to allow the **err_str** argument to be a single string, a list of strings, or a regex pattern (re.Pattern). 
2. The file I/O tests in tests/test_tester.py had to be updated to pass input and output file locations as absolute paths, because the Vivado working directory is not the same as the Python working directory.
3. tests/test_bidir.py and tests/test_hi_z.py skip the Vivado simulator since it does not support the **tran** Verilog keyword used in those tests (this issue is [briefly mentioned](https://forums.xilinx.com/t5/Simulation-and-Verification/unsupported-primitive/td-p/884456) on the Xilinx forums).

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -33,3 +33,25 @@ steps:
     label: "test"
     agents:
       fault: "true"
+  - command: |
+      # set up environment
+      source /etc/environment
+      echo $$PATH
+
+      # create virtual environment
+      python3.7 -m venv venv
+      source venv/bin/activate
+
+      # install python dependencies
+      pip install mantle pytest
+      pip install -e .
+      
+      # run tests
+      pytest tests
+
+      # deactivate virtual environment
+      deactivate
+    label: "fpga_verif"
+    timeout_in_minutes: 60
+    agents:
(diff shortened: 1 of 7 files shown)


sgherbst:
Vivado support is now tested using fpga-verif-* BuildKite agents -- updated the PR description to reflect this.

leonardt:
@sgherbst not high priority, but if you could get a chance, could you maybe just add a simple note to the README that we support vivado? We can extend the documentation as we get more interested users, but it would be good to at least notify interested parties that we can support it.

leonardt:
Come to think of it, we also could include some text on the support for the simulation targets that you added this summer, since that's also undocumented.

sgherbst:
Great points -- the documentation update is now implemented in PR #192

PR #188, opened 2019-12-12
Kratos debugging system integration

Kuree:
- Modified backend to handle both verilator and system-verilog outputs.
- Introduced a flag called `use_kratos` at top level.
- Added tests to test out loading runtime directly as well as a simple ALU test for both verilator and ncsim. The ALU is copied from the magma test.
- User needs to install `kratos_runtime` separately as it's not required to use fault.
- `kratos_runtime` currently only has native build for linux systems. I will try to figure out how to build it on macos later, but it's not high priority on my list since ncsim/vcs can't run on macos.

I will add `vcs` tests as a separate PR.

I think we should refactor `build_kratos_debug_info` as a utility function magma. I just copied it here, which should be refactored.

### DIFF
--- a/fault/system_verilog_target.py
+++ b/fault/system_verilog_target.py
@@ -45,7 +45,7 @@ class SystemVerilogTarget(VerilogTarget):
                  ext_libs=None, defines=None, flags=None, inc_dirs=None,
                  ext_test_bench=False, top_module=None, ext_srcs=None,
                  use_input_wires=False, parameters=None, disp_type='on_error',
-                 waveform_file=None, top_name=None, use_kratos=False):
+                 waveform_file=None, use_kratos=False):
         """
         circuit: a magma circuit
 
@@ -122,7 +122,7 @@ class SystemVerilogTarget(VerilogTarget):
 
         waveform_file: name of file to dump waveforms (default is
                        "waveform.vcd" for ncsim and "waveform.vpd" for vcs)
-        top_name: top level test bench name (default is {circuit_name}_tb
+        use_kratos: If True, set the environment up for debugging in kratos
         """
         # set default for list of external sources
         if include_verilog_libraries is None:
@@ -145,7 +145,7 @@ class SystemVerilogTarget(VerilogTarget):
         # call the super constructor
         super().__init__(circuit, circuit_name, directory, skip_compile,
                          include_verilog_libraries, magma_output,
-                         magma_opts)
+                         magma_opts, use_kratos=use_kratos)
 
         # sanity check
         if simulator is None:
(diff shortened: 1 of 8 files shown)


Kuree:
@leonardt Any idea why the buildkite failed? I don't think I ever touched these parts.

Kuree:
After this https://github.com/phanrahan/magma/pull/517 is approved, I can remove the duplicated code here.

fault/system_verilog_target.py
```diff
@@ -652,6 +669,15 @@ def input_wire(name):
...
         return f'{tabs*tab}{text}{nl}'
 
+    def link_kratos_lib(self):
+        from kratos_runtime import get_lib_path
+        lib_path = get_lib_path()
+        dst_path = os.path.join(self.directory, os.path.basename(lib_path))
+        if not os.path.isfile(dst_path):
+            os.symlink(lib_path, dst_path)
+        # also add the directory to the current LD_LIBRARY_PATH
+        self.sim_env["LD_LIBRARY_PATH"] = os.path.abspath(self.directory)
```
leonardt: I think this will need to do `DYLD_LIBRARY_PATH` if the platform is mac, if we haven't tested this yet, perhaps we can at least raise an error if we detect mac platform?

fault/verilator_target.py
```diff
@@ -105,6 +108,14 @@ def __init__(self, circuit, directory="build/",
...
         # Save settings
         self.disp_type = disp_type
+        self.use_kratos = use_kratos
+        if use_kratos:
+            try:
+                import kratos_runtime
+            except ImportError:
+                raise ImportError("Cannot find kratos-runtime in the system. "
+                                  "Please do \"pip install kratos-runtime\" "
+                                  "to install.")
```
leonardt: This runtime import logic is repeated in the system verilog target, can we move this to a shared helper function?

fault/verilator_target.py
```diff
@@ -534,6 +557,19 @@ def run(self, actions, verilator_includes=None, num_tests=0,
...
+        # if use kratos, symbolic link the library to dest folder
+        if self.use_kratos:
+            from kratos_runtime import get_lib_path
+            lib_name = os.path.basename(get_lib_path())
+            dst_path = os.path.abspath(os.path.join(self.directory, "obj_dir",
+                                                    lib_name))
+            if not os.path.isfile(dst_path):
+                os.symlink(get_lib_path(), dst_path)
+            # add ld library path
+            env = {"LD_LIBRARY_PATH": os.path.dirname(dst_path)}
```
leonardt: Same comment as above w.r.t. to macos (should either raise an error or we can just try setting DYLD_LIBRARY_PATH and hope that it works, travis does have a macos build so we could try running it on that).

Kuree: I don't have runtime build as a wheel for macos yet. I will debug with travis and once it's working I will and a separate PR.

PR #189, opened 2019-12-14
New SPICE simulation features

sgherbst:
This PR introduces some new features that are useful for SPICE simulation and fixes a few bugs.

Details:
1. A new **Action** called **GetValue** is added; it can be invoked using the **get_value(port)** method for a **Tester** object.  The **get_value** function returns an object with a **value** property that is assigned after the simulation completes.  This allows user to conveniently get data out of the simulator.  GetValue is currently supported for SystemVerilogTarget, VerilogAMSTarget, and SpiceTarget.
2. Since SPICE doesn't have a "connect by name" feature, we had been connecting assuming that ports were in alphabetical order.  This approach is now replaced by parsing the SPICE file using DeCiDa (the user can still fall back to using alphabetical order if needed).
3. Python traceback information is added when possible to error messages produced by **expect** actions.
4. Most parsing of SPICE results is now done by an external library, DeCiDa.
5. Load capacitors can now be added in SPICE simulations for more realistic simulations.
6. ngspice is added to the list of simulators tested in BuildKite regressions.
7. A few regression tests depended on the exact number of lines produced as output by simulator tools.  I found this was not reliable across different versions of **ncsim**, so I updated those tests to be a bit more flexible.


sgherbst:
I went ahead and added a GetValue action for SystemVerilog and VerilogAMSTarget as well, since this will be useful for **fixture**.  The PR is ready for review.

PR #190, opened 2019-12-17
Rework eval/step semantics

leonardt:
There are some issues with consistency between the verilator and system verilog simulation semantics.  This changes how step fundamentally works, but should enable better consistency going forward.

Basically, before `eval` was used to trigger an evaluation of the model in verilator, this doesn't really have a direct analog in system verilog.  Instead, SV TBs usually use the notion of a timestep to trigger events (e.g. move to the next timestep).

These changes deprecates the use of `eval` (it now calls step internally) and changes the semantics of `step` to indicate a step in the simulation time.  In verilator, `step` will call eval, and twiddle the clock if it exists.  In system verilog, `step` will advance the simulation time step, and the clock is implicitly twiddled each time step. 

In the future, we can add support for controlling the clock twiddle time step (e.g. the user can set the clock to twiddle every 5 time steps, allowing them to use step to evaluate combinational logic within a cycle).  This can be emulated in verilator by keeping track of the main simulation time and only twiddling the clock at the appropriate interval.  

This may break garnetflow since I see that it uses `eval` and `step`, I will see if I can come up with a fix or a way to gracefully introduce the new semantics (perhaps introduce a "synchronous tester" that adds these semantics)


leonardt:
It looks like this does break garnetflow, I'll re-open this with an alternative proposal for how to introduce these changes.

PR #191, opened 2019-12-17
Use non-blocking assign for SV poke

leonardt:
Verilator effectively uses non-blocking semantics because pokes are staged until `eval` is called.  This updates the system verilog target to do the same by using non-blocking assign.

### DIFF
--- a/fault/system_verilog_target.py
+++ b/fault/system_verilog_target.py
@@ -296,7 +296,7 @@ class SystemVerilogTarget(VerilogTarget):
         value = self.process_value(action.port, action.value)
         # Build up the poke action, including delay
         retval = []
-        retval += [f'{name} = {value};']
+        retval += [f'{name} <= {value};']
         if action.delay is not None:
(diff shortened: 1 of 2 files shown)


Kuree:
I don't think the semantics is correct. If all the pokes are non-blocking, then the values will be updated in the NBA region, but a lot of semantics such as DPI calls and `$display` is evaluated in the active region. I won't recommend to force the evaluation behavior in SV simulator.

Notice that Verilator is *not* SV spec-compliant by choice. That means the test bench is not setup the same way as SystemVerilog test bench files. Matching a spec-compliant simulator to a non-compliant one seems to be an odd choice. Maybe I'm missing something here?

You can refer to Figure 4-1 in ISO 1800-2017.

leonardt:
The goal of fault is to provide an abstraction over both simulators, otherwise you wouldn't be able to write a generic test bench and target both simulators, in order to do this, the semantics for the actions need to be consistent across the generated test benches.

PR #193, opened 2019-12-21
Implement GetValue for Verilator + Code Refactoring

sgherbst:
This PR implements the GetValue action for Verilator as is already supported by SystemVerilogTarget, VAMSTarget, and SpiceTarget.  While I was at it, I also refactored some common code in SystemVerilogTarget and VerilatorTarget into VerilogTarget.

Details:
1. Implemented **GetValue** for **VerilatorTarget** in the same way it is implemented for **SystemVerilogTarget**: namely, values are written to an external text file during simulation, then read back to populate the **value** variable of the **GetValue** objects.  The post-processing code now lives in **VerilogTarget** since it is common to both **SystemVerilogTarget** and **VerilatorTarget**.
2. Refactored **make_if**, **make_while**, and **make_loop** from **SystemVerilogTarget** and **VerilatorTarget** into **VerilogTarget**.  These functions all make use of the new **make_block** function in VerilogTarget, which implement an indented code block using language-specific block-start / block-stop strings.
3. Assignments and variable declarations in **SystemVerilogTarget** are now handled through **add_assign** and **add_decl**, respectively, which have checks to avoid re-declarations and re-assignments.
4. Added support for string and list arguments to **generate_action_code** in **VerilogTarget**, which makes it easier to mix hand-written lines of code with actions like **If** and **Loop** (see for example **make_file_read** in either SystemVerilogTarget or VerilatorTarget).  This in turn allows more of the testbench code to be generated with existing actions.
5. On a similar note, opening and closing the file used for **GetValue** actions is now implemented using existing **FileOpen** and **FileClose** actions; the presence of one or more **GetValue** actions causes this file I/O actions to be wrapped around the user-provided **actions** list.
6. Added an optional "count" parameter to the **Loop** action that allows the user to specify whether the loop should count up or count down.  This made it easier to implement the handling of endianness in **make_file_read** and **make_file_write** using a **Loop** action.


sgherbst:
This PR is now up-to-date with the latest changes to **master** & ready for review.  As a side note, I think it may be easier to appreciate refactoring in **fault/verilator_target.py** and **fault/system_verilator_target.py** by directly viewing the new files rather than the diff, due to GitHub's formatting.

PR #194, opened 2020-01-06
Support bitwise assignments

leonardt:
This adds support for assigning to individual bits in an array of bits, e.g.
```
# defined in io of DUT
IO = ["myBits", m.In(m.Bits[2])]

# in fault tester
test.circuit.myBits[0] = 1
```

In verilator, this is done in C using bit twiddling to clear the bit index, then or in the poke value for the bit.  In system verilog this simply uses the standard index assignment logic.  This requires an update to the name generation logic when we encounter a reference to an index in an array of bits:
* in C, don't emit the index for the array element, instead emit the containing array, and the poke logic handles the bit assignment by replacing the original value with the bit twiddled value
* in system verilog, emit the index as a subscript `x[{index}]` to refer to the desired bit position (normally we assume array indices are flattened out into the name, for general recursive arrays).


fault/verilator_target.py
```diff
@@ -252,6 +277,7 @@ def make_poke(self, i, action):
                                    i in range(value.file.chunk_size))
                 value = f"({value}) & 0x{mask}"
             value = self.process_value(action.port, value)
+            value = self.process_bitwise_assign(action.port, name, value)
```
rsetaluri: Just confirming this will work for nested arrays? Like `tester.circuit.I[0][0] = 1` where `I` is `Array[Bits[2], 2]`? I think the line above will give us the flattened thing?

fault/verilator_target.py
```diff
@@ -341,6 +367,9 @@ def make_expect(self, i, action):
             return asserts
         else:
             value = self.process_value(action.port, value)
+            actual_value = f"top->{name}"
```
rsetaluri: nit: rename `actual_value` to `port_value`

fault/verilog_utils.py
```diff
@@ -6,7 +6,10 @@ def verilog_name(name):
         return str(name)
     if isinstance(name, m.ref.ArrayRef):
         array_name = verilog_name(name.array.name)
-        return f"{array_name}_{name.index}"
+        if isinstance(name.array.T, m._BitKind):
+            return f"{array_name}[{name.index}]"
+        else:
```
rsetaluri: nit: get rid of else

fault/verilator_target.py
leonardt: Those should be flattened out by this point (since coreir flattens out arrays down to arrays of bits), but I'll add a test to be sure

PR #195, opened 2020-01-07
speed up travis build

Kuree:
Use pre-built wheel for z3. This should speed up the macos build significantly.


leonardt:
Merged into https://github.com/leonardt/fault/pull/196

PR #196, opened 2020-01-07
Use new coreir wheel

leonardt:


Kuree:
Looks like macos wheel is broken somehow. I don't have access to a mac machine. Can you take a look at it tomorrow?

Kuree:
@leonardt Reading from the log it seems like the library path in macos wheel is wrong. Usually on linux I would manually inspect the shipped libraries with `ldd` and see which library is missing.

On linux it is fixed with the official `auditwheel` tool by providing library locations with `LD_LIBRARY_PATH`. I think you could try this tool on macos: https://pypi.org/project/delocate/,
which offers similar ability as `auditwheel`.

PR #197, opened 2020-01-10
Add support for poke/expect slice

leonardt:
Adds the ability to poke/expect on a slice of bits, e.g.
```
    tester.circuit.I[1:] = 0b01
    tester.eval()
    tester.circuit.O[1:].expect(0b01)
```

### DIFF
--- a/fault/tester.py
+++ b/fault/tester.py
@@ -162,7 +162,8 @@ class Tester:
         if isinstance(port, SelectPath):
             if self.is_recursive_type(port[-1]):
                 return recurse(port[-1])
-        elif self.is_recursive_type(port):
+        elif not isinstance(port, fault.WrappedVerilogInternalPort) and \
+                self.is_recursive_type(port):
             return recurse(port)
 
         if not isinstance(value, (LoopIndex, actions.FileRead,
@@ -203,7 +204,8 @@ class Tester:
         if isinstance(port, SelectPath):
             if self.is_recursive_type(port[-1]):
                 return recurse(port[-1])
-        elif self.is_recursive_type(port):
+        elif not isinstance(port, fault.WrappedVerilogInternalPort) and \
+                self.is_recursive_type(port):
             return recurse(port)
 
         # set defaults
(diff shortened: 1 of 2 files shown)


fault/tester.py
```diff
@@ -126,8 +127,11 @@ def make_target(self, target: str, **kwargs):
         raise NotImplementedError(target)
 
     def is_recursive_type(self, T):
+        if isinstance(T, fault.WrappedVerilogInternalPort):
+            return False
         return isinstance(T, m.TupleType) or isinstance(T, m.ArrayType) and \
-            not isinstance(T.T, (m._BitKind, m.BitType))
+            not isinstance(T.T, (m._BitKind, m.BitType)) or \
```
cdonovick: This seems like a pretty nasty condition.   Is this something we could push into magma to simplify life?

cdonovick: I feel like any time a type is added this will have to change.  Which is bad.

leonardt: I don't know if there's a way around it.  Fault needs to understand recursive types.  If a type subclasses the recursive types, that should work, but if a new type is added, fault has no way of knowing how to handle it unless we can define some generic traversal function.  I think the way to avoid the condition would be some way to generically visit a recursive type.

cdonovick: > I think the way to avoid the condition would be some way to generically visit a recursive type.
 
In hwtypes adts define `field_dict` to make this type of recursion easier.   Maybe magma could have a similar sort of protocol?

leonardt: Sure, but I'd separate those changes out from this as this adds the basic functionality, while that involves a refactor of a lot of internal code to adhere to the new interface

cdonovick: Of course.  This was more of an observation than a request for code changes.  I generally dislike complicated if conditions because they break and are generally a sign of poor code structuring.  Regardless, neither magma nor fault are mine; doesn't matter to me if *you* have to write a million ifs.

PR #199, opened 2020-01-11
Raise attribute error instead of exception

leonardt:
This adjusts the code so that it will raise an AttributeError (implicitly) rather than explicitly raising a plain Exception so `hasattr` will work.


leonardt:
Side effect, this also relaxes the check so it allows getting attributes of non tuples.

PR #205, opened 2020-01-27
Add coverage command to verilator and ncsim

Kuree:
This maybe a duplicate of #65.

What does this PR do:
- Add flag `coverage` to all targets
- Tested on `ncsim` and `Verilator`. Haven't tried `vcs` yet since I'm not an expert in `vcs`; someone needs to help me out there.
- For `Verilator`, the coverage file can be parsed directly, whereas for `ncsim` we need `IMC` tools since it's a gzipped file linked to its compiled model. `kratos` has built-in ability to parse both format. Maybe there is an opportunity for tight integration so that fault can digest coverage information through `kratos`.


hofstee:
In VCS I think you can do `vcs -cm line` but it's been a while since I've done anything of the sort.

PR #210, opened 2020-02-20
Add skip run flag

rsetaluri:


leonardt:
Hmm, it seems like we may want to factor it such that `compile` will actually generate the test bench collateral instead of `run`, so we could achieve this with just using `compile` rather than `compile_and_run`, but this get's the job done for now.

rsetaluri:
Was going to try this but actually realized it's not possible that way, because the test bench requires the actions which are only available at `run` time, not "compile" (`__init__`) time.

PR #211, opened 2020-02-21
Add an option to use SVA

Kuree:
This can be used together with properties etc.

I don't have access to OSX so I have no idea how to fix brew. It seems like the OSX setup somehow is outdated/wrong, although I didn't change anything with the travis config.

### DIFF
--- a/fault/system_verilog_target.py
+++ b/fault/system_verilog_target.py
@@ -927,6 +927,7 @@ class SystemVerilogTarget(VerilogTarget):
         cmd += self.def_args(prefix='+define+')
 
         # misc flags
+        cmd += ['-access', '+rwc']
         cmd += ['-sverilog']
         cmd += ['-full64']
         cmd += ['+v2k']
--- a/fault/system_verilog_target.py
+++ b/fault/system_verilog_target.py
@@ -56,7 +56,8 @@ class SystemVerilogTarget(VerilogTarget):
                  ext_libs=None, defines=None, flags=None, inc_dirs=None,
                  ext_test_bench=False, top_module=None, ext_srcs=None,
                  use_input_wires=False, parameters=None, disp_type='on_error',
-                 waveform_file=None, coverage=False, use_kratos=False):
+                 waveform_file=None, coverage=False, use_kratos=False,
+                 use_sva=False):
         """
         circuit: a magma circuit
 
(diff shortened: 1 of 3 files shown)


sgherbst:
Looks like this is a brand new bug in Travis macOS builds: https://travis-ci.community/t/macos-build-fails-because-of-homebrew-bundle-unknown-command/7296/14.  I added the workaround suggested there and will see if it fixes the problem.  Once the issue is fixed by the Travis team, we should remove the extra "update: true" option since it increases the macOS build time.

Kuree:
@sgherbst Thanks!

fault/system_verilog_target.py
```diff
@@ -860,7 +868,6 @@ def ncsim_cmd(self, sources, cmd_file):
         cmd += self.def_args(prefix='+define+')
 
         # misc flags
-        cmd += ['-access', '+rwc']
```
leonardt: Is this okay to turn off in general? I don't remember why this option is included.  @sgherbst do you use this option ever? I think this may have been copied over from an old ncsim script that Nikhil gave me as an example, so I'm not sure if it's actually used by anyone.

Based on this https://www.edaboard.com/showthread.php?23842-What-is-the-access-rwc-option-in-NCSIM it seems that this enables access to objects in the design.  We might need this for working with internal signals? (But the tests still pass so that may not be the case)

sgherbst: I think “access -r” is needed if you want to dump waveforms, so we would want to enable at least that if the user has requested waveform probing (I sort of remember trying without in the past and getting an empty waveform database).  However, this option shouldn’t be added when not needed because it slows down simulation.  

“access -rw” further enables write access to signals from TCL, and “access -rwc” add connectivity access for driver tracing for debugging.  These options can be useful, but further slow down simulation.

Kuree: `-rwc` is needed if you're using VPI or TCL related simulation where the runtime library/scripts needs access to the simulation objects, especially the ability to override values. This is unrelated to the signal access in test benches since the access is known during compilation. As Steven said, this will slow down simulation simulation.

I think `-access r` is required to dump the VCD files, which I forgot to check in this PR.

Kuree:
@leonardt I will revert the changes to the flags and submit a separate PR that fixes the flags.

leonardt:
sounds good, thanks!

PR #217, opened 2020-02-27
update magma IO syntax

hofstee:


hofstee:
How do you append to IO with the new syntax?

leonardt:
There's an open issue on this here: https://github.com/phanrahan/magma/issues/596

For now, the best way would be to incrementally construct a python dictionary ("appending" by adding keys to the dictionary), then passing the dict to IO

leonardt:
Another option is to keep `IO` as a list and do a
```
io = m.IO(**{key: value for key, value in zip(IO[::2], IO[1::2])})
```

but i think this will trigger a different warning (mixing IO and io), so ideally you'd also rename the IO list to be something like `IO_`

leonardt:
Manually fixing up these cases might be okay though, if writing the script is too tricky

leonardt:
(Let me know if that's the case and I can go ahead and take care of the manual fixes)

hofstee:
I think these might need to be manually handled. (Maybe you could still do `io += m.ClkInterface()`?)

PR #218, opened 2020-03-02
Add tester call method

leonardt:


fault/tester.py
```diff
@@ -536,6 +536,26 @@ def internal(self, *args):
         # return a SelectPath containing the desired path
         return SelectPath([self.circuit] + list(args))
 
+    def __call__(self, *args, **kwargs):
```
rsetaluri: Feel like the interface to this function can just be to return the references to the outputs, not to a `peek`. That avoids having to modify the `Peek` class interface. Adding a back-reference to a `Tester` object and a `expect()` function on a `Peek` instance seems a little out of place and clunky (and avoids making an breaking API change).

We could even return the "view" on those ports so you can still call `expect` on them:

```python
tester.circuit.O.expect(x)
tester(a, b).expect(y)
```

(A third option would be to **extend** the `Expect` class interface to take in a `Peek` object (instead of a port), but given that we have the port-view's as above, this seems suboptimal.)

fault/tester.py
```diff
@@ -536,6 +536,26 @@ def internal(self, *args):
...
+        Poke the inputs of the circuit using *args (ordered, anonymous input
+        reference, excluding clocks) and **kwargs.  **kwargs will overwrite any
+        inputs written by *args.
+
+        Evaluate the circuit
+
+        Return the "peeked" output(s) of the circuit (tuple for multiple
+        outputs)
+        """
+        for arg, port in zip(args, self._circuit.interface.outputs()):
```
rsetaluri: I don't see `kwarg` override happening here?

leonardt: oops forgot it! heh

fault/tester.py
```diff
@@ -536,6 +536,28 @@ def internal(self, *args):
...
```
rsetaluri: Also this logic seems a bit weird...it expects `args` to be of length of # of inputs, then allows overriding using `kwargs`? Seems the expectation should be that `len(args) + len(kwargs) == num_inputs`...or just have a check that `len(args) <= num_inputs` and wire up only the first `m` inputs. E.g. see https://github.com/phanrahan/magma/blob/5bdb58bda377b363824629b719ddb0a4fb433fa8/magma/circuit.py#L391; having parity with that func. seems reasonable.

leonardt: zip will truncate by default, so you can use less inputs and it will skip the extra off the end (e.g. if they are wired up by kwargs

leonardt: I can add check that the total sums up to the expected amount, otherwise issue a warning

leonardt: The weird thing is, you can skip setting the clock, since it might be combinational evaluation, so actually that check might be overly conservative again

leonardt: So really it should be len(args) + len(kwargs that are not clock args)

fault/tester.py
leonardt: I'm not sure if it's worth the extra complexity?

I think doing something like
```
tester.peek(circuit.O).expect(3)
```
doesn't seem to unreasonable of thing to support.

In fact, you can already write
```
tester.circuit.O.expect(3)
```
but the wrapper machinery in the setattr interface already handles this, so it could be simplified if a peek object supported expect.

rsetaluri: My main point is that the actions should be generic objects that are independent of anything to do with the Tester class, which is why they're defined as such, and the Tester class consumes actions and does stuff with them.

Can't we get the same functionality out of call with returning the view object? So

```python = 
result = tuple(getattr(self.circuit, name) for name in self._circuit.interface.inputs_by_name())
```

and then we still have the same syntax in the calling code:

```python
    tester = fault.Tester(AndCircuit)
    for i, j in zip(range(2), range(2)):
        tester(i, j).expect(i & j)
```

since `tester(i, j)` returns a port-view object

rsetaluri: (Perhaps my first comment was ambiguous, by "view" object I mean the existing view objects. The fact that we can already do `tester.circuit.O.expect(3)` is exactly what I'm suggesting we leverage)

fault/tester.py
rsetaluri: seems reasonable, the only thing is to avoid user confusion if either (a) a kwarg overrides an arg, or (b) a wrong number of total args are passed in (too few or too many)

PR #219, opened 2020-03-02
Interactive tester

leonardt:
This refactors the tester class to use an abstract class pattern, and adds a new concrete implementation in the form of an InteractiveTester.

This is a thin wrapper over the magma Python simulator, mainly to handle the conversion of types going in and out of the simulator (using our standard BitVector, Bit pattern now).

There's improvements to be made (such as changing the magma simulator target to reuse this logic in the staged tester), but this is aimed at getting the minimum functionality required for the magmathon.


fault/magma_utils.py
```diff
@@ -0,0 +1,6 @@
+import magma as m
+
+
+def is_recursive_type(T):
+    return issubclass(T, m.Tuple) or issubclass(T, m.Array) and \
```
rsetaluri: going forward let's try to stick to this for long lines: http://google.github.io/styleguide/pyguide.html#3-python-style-rules, in particular 

> Do not use backslash line continuation except for `with` statements requiring three or more context managers.

So

```python
return (issubclass(T, m.Tuple) or issubclass(T, m.Array) and
        not issubclass(T.T, m.Digital))
```

fault/tester/abstract_tester.py
```diff
@@ -0,0 +1,231 @@
+import magma as m
+from abc import abstractmethod
+from ..wrapper import CircuitWrapper
+from ..select_path import SelectPath
+from ..wrapped_internal_port import WrappedVerilogInternalPort
+from ..magma_utils import is_recursive_type
+
+
+class AbstractTester:
```
rsetaluri: `AbstractTester` -> `TesterBase` (only because `AbstractTester` could sound like `SymbolicTester` which is an actual thing :D here)

fault/tester/abstract_tester.py
```diff
@@ -0,0 +1,231 @@
...
+        # break existing TBs)
+        self.clock_initialized = False
+        # Only report once, in case the user calls step with an uninitialized
+        # clock many times
+        self.clock_init_warning_reported = False
+        if reset is not None and not isinstance(reset, m.Reset):
+            raise TypeError(f"Expected reset port: {reset, type(reset)}")
+        self.reset_port = reset
+
+    def get_port_type(self, port):
```
rsetaluri: needs to be a member function or can be helper?

fault/tester/abstract_tester.py
```diff
@@ -0,0 +1,231 @@
...
+        if isinstance(port, SelectPath):
```
rsetaluri:
```python
        if isinstance(port, SelectPath):
            port = port[-1]
        if isinstance(port, WrappedVerilogInternalPort):
            return port.type_
        return type(port)
```

fault/tester/interactive_tester.py
```diff
@@ -0,0 +1,115 @@
...
+        # works
+        expected = expected.as_bool_list()
+    if isinstance(port, m.Array):
+        for i in range(port.N):
+            check(got[i], port[i], expected[i])
+        return
+    assert got == expected, f"Got {got}, expected {expected}"
+
+
+class PythonTester(AbstractTester):
```
rsetaluri: Should this inherit from `InteractiveTester`?

fault/tester/interactive_tester.py
```diff
@@ -0,0 +1,115 @@
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.simulator = PythonSimulator(self._circuit, self.clock)
+
+    def eval(self):
+        self.simulator.evaluate()
+
+    def process_port(self, port):
```
rsetaluri: needs to be a member function or can be private helper?

fault/tester/interactive_tester.py
```diff
@@ -0,0 +1,115 @@
...
+        check(got, port, expected)
+
+    def peek(self, port):
+        return self._get_value(port)
+
+    def print(self, format_str, *args):
+        got = [self._get_value(port) for port in args]
+        values = ()
+        for value, port in zip(got, args):
+            if isinstance(port, m.Array) and \
```
rsetaluri: same re: long lines

fault/tester/interactive_tester.py
```diff
@@ -0,0 +1,115 @@
...
+                    issubclass(port.T, m.Digital):
+                value = BitVector[len(port)](value).as_uint()
+            elif isinstance(port, m.Array):
+                raise NotImplementedError("Printing complex nested"
```
rsetaluri:
```python
                raise NotImplementedError("Printing complex nested "
                                          "arrays")
```

fault/tester/staged_tester.py
```diff
@@ -127,71 +113,25 @@ def make_target(self, target: str, **kwargs):
...
-                    isinstance(port[-1].name, m.ref.AnonRef))):
-                return recurse(port[-1])
-        elif self.is_recursive_type(type(port)):
-            return recurse(port)
-        elif not isinstance(port, WrappedVerilogInternalPort) and\
-                isinstance(port.name, m.ref.AnonRef):
-            return recurse(port)
+        recursed = super().poke(port, value, delay)
+        if recursed:
+            return
```
rsetaluri: should this also return a `bool` so that `Tester::poke` and `AbstractTester::poke` have the same interface? i.e. `return True` here and `return False` at end of func.

fault/verilator_target.py
```diff
@@ -187,6 +187,12 @@ def process_value(self, port, value):
             return value
         elif isinstance(value, actions.Var):
             return value.name
+        elif isinstance(value, actions.FileRead):
+            mask = "FF" * value.file.chunk_size
+            value = " | ".join(f"{value.file.name_without_ext}_in[{i}]" for
```
rsetaluri:
```python
            value = " | ".join(f"{value.file.name_without_ext}_in[{i}]"
                               for i in range(value.file.chunk_size))
```

fault/verilator_target.py
```diff
@@ -187,6 +187,12 @@ def process_value(self, port, value):
...
+                               i in range(value.file.chunk_size))
+            value = f"({value}) & 0x{mask}"
```
rsetaluri: `return f"({value}) & 0x{mask}"`

fault/verilator_target.py
```diff
@@ -187,6 +187,12 @@ def process_value(self, port, value):
             return value
         elif isinstance(value, actions.Var):
             return value.name
+        elif isinstance(value, actions.FileRead):
```
rsetaluri: this (and all `elif`'s above) can be `if` due to early exits

fault/tester/abstract_tester.py
leonardt: Yes, the children inherit this helper and refer to it

fault/tester/staged_tester.py
leonardt: It's not meant to be part of the external interface, it's just to reuse the recursive logic provided by the base class.

rsetaluri: Got it. In that case would recommend doing the following:

In `TesterBase`:

```python
def _poke(...):
    # Whatever logic was originally in poke() before.

def poke(...):
    self._poke(...)
```

Then in this class `Tester` override `poke()` and refer to `super()._poke(...)`. That way `poke()` has the same interface across and `super._poke()` can be reused.

Not a big deal either way, just seems like a cleaner/more consistent interface.

fault/tester/base.py
```diff
@@ -0,0 +1,282 @@
...
+        # break existing TBs)
+        self.clock_initialized = False
+        # Only report once, in case the user calls step with an uninitialized
+        # clock many times
+        self.clock_init_warning_reported = False
+        if reset is not None and not isinstance(reset, m.Reset):
+            raise TypeError(f"Expected reset port: {reset, type(reset)}")
+        self.reset_port = reset
+
+    def get_port_type(self, port):
```
rsetaluri: Got it. However, if it's not meant to be a part of the external interface, and it doesn't modify or use any member variables, then would suggest it being an helper function external to the class. It can still be used by child classes, or whoever. But not a big deal either way.

PR #222, opened 2020-03-10
Init synchronous tester

leonardt:
This adds an initial version of the synchronous tester.  This removes the ability to call `eval` and `step` directly, and instead presents an `advance_cycle` interface (which call step internally).

During code generation, `step` translates to `#5` (where 5 is parametrizable) which corresponds to a half-cycle.  `advance_cycle` always advances two steps (so you get a `#10` out).  The synchronous tester requires a clock, and this clock is used to generate an `always #5 CLK = ~CLK`.  Again the delay is parametrizable.  In the future, we could add support for multiple, parallel clock drivers, but I don't think this is an immediate use case.

The verilator target doesn't make any changes to the code generation currently because I think the interface should satisfy the desired behavior, but we can update it we find any semantic discrepancies.

The main change here from the original tester is disallowing `eval` (which relied on pound delay to implement) so now the output waveform should be consistent with the clock ticks.  When using `step` in verilator, it similarly will only advance simulator time in the waveform once during step (so you don't get the intermediated poke eval time steps).

### DIFF
--- a/fault/tester/staged_tester.py
+++ b/fault/tester/staged_tester.py
@@ -422,3 +422,6 @@ class IfTester(Tester):
 
     def _else(self):
         return ElseTester(self.else_actions, self.circuit, self.clock)
+
+
+StagedTester = Tester
--- a/fault/tester/synchronous.py
+++ b/fault/tester/synchronous.py
@@ -1,5 +1,6 @@
 from .staged_tester import StagedTester
 from ..system_verilog_target import SynchronousSystemVerilogTarget
+from ..verilator_target import SynchronousVerilatorTarget
 
 
 class SynchronousTester(StagedTester):
@@ -20,4 +21,7 @@ class SynchronousTester(StagedTester):
         if target == "system-verilog":
             return SynchronousSystemVerilogTarget(self._circuit,
                                                   clock=self.clock, **kwargs)
+        elif target == "system-verilog":
+            return SynchronousVerilatorTarget(self._circuit, clock=self.clock,
+                                              **kwargs)
         return super().make_target(target, **kwargs)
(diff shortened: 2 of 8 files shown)


leonardt:
Ok, I found the semantic inconsistency with the verilator/system verilog targets for these synchronous cases.  From the verilator docs: https://www.veripool.org/projects/verilator/wiki/Manual-verilator#EVALUATION-LOOP
> When eval() is called Verilator looks for changes in clock signals and evaluates related sequential always blocks, such as computing always_ff @ (posedge...) outputs. Then Verilator evaluates combinatorial logic.

The issue is that when we stage a set of pokes, we call eval() before updating the clock value.  This causes combinational logic to trigger before the clock tick occurs.  The expectation (following the behavior of system verilog non blocking) is for the clock tick to occur first.  This is already implemented by the verilator `eval` semantics (noted above) but we shouldn't call `eval` on the pokes for the current tilmestep before the clock value is updated.

fault/tester/staged_tester.py
```diff
@@ -422,3 +422,6 @@ def __init__(self, circuit: m.Circuit, clock: m.Clock = None):
 
     def _else(self):
         return ElseTester(self.else_actions, self.circuit, self.clock)
+
+
+StagedTester = Tester
```
rsetaluri:
```python
class StagedTester(Tester):
    pass
```
instead? unless we need `is` relation

fault/tester/synchronous.py
```diff
@@ -0,0 +1,27 @@
...
+        raise TypeError("Cannot eval with synchronous tester")
+
+    def advance_cycle(self):
+        self.step(2)
+
+    def make_target(self, target, **kwargs):
+        if target == "system-verilog":
+            return SynchronousSystemVerilogTarget(self._circuit,
+                                                  clock=self.clock, **kwargs)
+        elif target == "system-verilog":
```
rsetaluri: elif -> if

rsetaluri: nit: elif -> if

fault/tester/staged_tester.py
leonardt: I think we should eventually deprecate `Tester` and have `StagedTester` be the default name since it's more explicit (compared to interactive Tester) so it's not really meant to be a subclass, rather an alias

PR #230, opened 2020-04-02
Add flatten method to real_type

standanley:
This is a tiny change to fix real_type after a recent change to magma. I can't run all the tests on my machine since I don't have all the simulators installed, but it did fix some things at least.


standanley:
Replaced by #231

PR #233, opened 2020-04-11
Test pe data gate enhancements

rsetaluri:
Few enhancements needed for testing garnet PE data gating. think they should be robust and generally good.


rsetaluri:
Keeping branch around as garnet needs it still. Making a release first.

PR #235, opened 2020-04-28
Support poking protocol values

leonardt:
Adds initial support for https://github.com/cdonovick/peak/issues/154#issuecomment-618686502

### DIFF
--- a/tests/test_magma_protocol.py
+++ b/tests/test_magma_protocol.py
@@ -7,23 +7,15 @@ from hwtypes import BitVector
 def test_foo_type_magma_protocol():
     class FooMeta(m.MagmaProtocolMeta):
         def _to_magma_(cls):
-            # Need way to retrieve underlying magma type
             return cls.T
 
         def _qualify_magma_(cls, direction: m.Direction):
-            # Need way to create a new version (e.g. give me a Foo with the
-            # underlying type qualified to be an input)
             return cls[cls.T.qualify(direction)]
 
         def _flip_magma_(cls):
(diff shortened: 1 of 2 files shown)


tests/test_magma_protocol.py
```diff
@@ -0,0 +1,62 @@
+from typing import Optional
+import magma as m
+import fault
+from hwtypes import BitVector
+
+
+def test_foo_type_magma_protocol():
+    class FooMeta(m.MagmaProtocolMeta):
+        def _to_magma_(cls):
+            # Need way to retrieve underlying magma type
```
rsetaluri, line 10: can we kill these comments (they seem to be duplicated from the MagmaProtocolMeta definition)

tests/test_magma_protocol.py
```diff
@@ -0,0 +1,62 @@
...
+        def _get_magma_value_(self):
+            # Need way to access underlying magma value
+            return self._val
+
+        def non_standard_operation(self):
+            v0 = self._val << 2
+            v1 = m.bits(self._val[0], len(self.T)) << 1
+            return Foo(v0 | v1 | m.bits(self._val[0], len(self.T)))
+
+    @m.circuit.sequential
```
rsetaluri, line 47: prefer that we don't use sequential to test things that don't have to nec. do with sequential. It makes debugging unnec. harder.

Presumably this could just be

```python
class Bar(m.Circuit):
    io = m.IO(foo=m.In(Foo[m.Bits[8]]), O=m.Out(m.Bits[8]))
    io.O @= io.foo.non_standard_operation()
```

PR #237, opened 2020-05-05
Fix deadlock in subprocess logic

leonardt:
See
https://stackoverflow.com/questions/1445627/how-can-i-find-out-why-subprocess-popen-wait-waits-forever-if-stdout-pipe
for some more details

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ matrix:
     - CC=gcc-7
     - CXX=g++-7
   - os: osx
-    osx_image: xcode11.4
+    osx_image: xcode9.4
     addons:
       homebrew:
         packages:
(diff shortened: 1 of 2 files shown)


Kuree:
@leonardt Do you want me to try out Github actions instead of Travis? Github actions uses latest OSX.

leonardt:
@Kuree sure, but I'm not exactly sure if this issue is fixed with the latest OS X, looking into it now, I think perhaps the verilator version on homebrew got updated and now it's not working with the travis environment

Kuree:
My guess is that the apple clang came with the OSX image is too old. Switching to newest image should help.

Kuree:
@leonardt See https://github.com/leonardt/fault/pull/238, which fixes the macos issue via Github Actions.

PR #238, opened 2020-05-06
Github CI

Kuree:
This should replace Travis due to lack of maintenance on their side. Using Github CI should also significantly speed up the tests as well. This should fix the macos failing issue on Travis.

Haven't written CI script for deployment yet, but should be easy to do so.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -1,0 +1,45 @@
+name: Linux Test
+
+on: [push]
+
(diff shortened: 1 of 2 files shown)


leonardt:
Can you merge/rebase master so you get the new travis script with OS X skipped? Let's keep travis for now until we update the deploy logic (once we migrate that to this flow, we can drop the travis script)

Issue #240, opened 2020-06-11
[RFC] Fault Property Language


leonardt:
# Design Goals
* Provide a python embedded language for writing assert/cover properties using magma values
* Simple integration with magma design code (write inline with generators) to minimize overhead
* Compile through CoreIR to SystemVerilog to integrate with mainstream verification tools

# Initial Requirements
These are the initial features identified by users.  The goal is to provide a minimal set of useful primitives that hopefully capture the various issues that might arise in the design of the language.  These primitives should be immediately useful for writing assertions, without being too complex to facilitate rapid development of an initial prototype (e.g. we don't need to support all of SVA from the start).

## Types of assertions
1. **Immediate/final assertions** - unclocked boolean expressions that are either applied in the `initial` or `final` phase of simulation.
    ```python
    f.assert_immediate(prop0, name="name0")
    f.assert_final(prop1, name="name1")
    ```
2. **Concurrent properties** - clocked (and usually guarded by reset).  Three main types: `assert`, `covers`, and `assume`.  `assume` is use for formal property verification and is treated as an `assert` in simulation.
   ```python
   f.assert(prop0, name="name0", on=m.posedge(io.CLK)).disable_if(!io.RESETN)

   f.cover(prop1 , name="name1", on=m.posedge(io.CLK)).disable_if(!io.RESETN)

   f.assume(prop2, name="name2", on=m.posedge(io.CLK)).disable_if(!io.RESETN)
   ```
   `disable_if` is optional (maybe just make it kwarg to be consistent?).  Should support negedge, as well as multiple clocks.

## Property Primitives
We present two proposed interfaces.  The first uses a "hack" to define custom infix operators (see http://tomerfiliba.com/blog/Infix-Operators/ for an example).  This provides a decently ergonomic alternative to SVA's convenient operators given Python's lack of corresponding operator definitions or ability to define custom operators.  The second uses a string interface to allow the user to pass an operator as string.  This is mainly targeted towards experienced SVA users that would like to use their familiar syntax.
1. **Suffix Implication** - `S` is the antecedent (triggering condition), `P` is the consequent (checked when `P` holds).  SVA's *overlapping implication* operator `S |-> P`.
   ```python
   S |f.implies| P
   f.sva(S, "|->", P)
   ```
2. **Pound Delay** - Basic primitive for defining temporal sequences of boolean expressions that are evaluated over time (clock ticks).  We use square braces for consistency with variable pound delay (next item) which can leverage slice syntax.
   ```python
[...]

PR #241, opened 2020-07-01
Property Prototype

leonardt:
Initial prototype of https://github.com/leonardt/fault/issues/240

Seems like we should be able to pull this off.  Will also add an example of the sva syntax.

For now we just compile to inline_verilog, eventually we should add corresponding coreir primitives and compile to that representation.

Unforuntately, verilator does not support the concurrent assertion syntax (at least the ## operator was causing an error. (also see https://www.veripool.org/projects/verilator/wiki/Manual-verilator#Assertions).

Fortunately, this gives us a good reason to support compiling these to pono (next gen CoSA, CC @makaimann) to test this feature in the open source environment.

There's still some work to organize the property code (e.g. class hierarchy and reuse for the infix operator logic, as well as the compiler logic) but this was just a first stab at showing that it's possible.  I started by defining a compile method on the property subtypes, but I found passing around the format_kwargs dictionary annoying, so I think it makes more sense to define a Compiler visitor (ala ast NodeTransformer pattern) so the compile code will live inside a separate class rather than with the properties.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -16,7 +16,7 @@ steps:
       # install python dependencies for testing
       pip install pytest-cov pytest-codestyle
       pip install --upgrade "mantle>=2.0.0"
-      pip install vcdvcd
+      pip install vcdvcd decorator
 
       # install fault
       pip install -e .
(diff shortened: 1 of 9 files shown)


leonardt:
Depends on https://github.com/phanrahan/magma/pull/760

PR #242, opened 2020-07-08
Fix clock init bug

leonardt:
The default clock logic was being inherited by sub-testers (e.g. for if/while loops) which was causing an unexpected setting of the clock to zero inside the TB

### DIFF
--- a/fault/tester/staged_tester.py
+++ b/fault/tester/staged_tester.py
@@ -79,12 +79,15 @@ class Tester(TesterBase):
         """
         super().__init__(circuit, clock, reset, poke_delay_default,
                          expect_strict_default)
-        if clock is not None:
-            # Default initalize clock to zero (e.g. for SV targets)
-            self.poke(clock, 0)
+        self.init_clock()
         self.targets = {}
         # For public verilator modules
         self.verilator_includes = []
+
+    def init_clock(self):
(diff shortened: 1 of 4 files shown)


fault/tester/staged_tester.py
```diff
@@ -393,7 +396,16 @@ def __str__(self):
         return self.name
 
 
-class LoopTester(Tester):
+class NoClockInit:
```
rsetaluri, line 399: Seems like it would be better to have `TesterBase` class have a default implementation of `def init_clock(self): pass`. And potentially have a `SynchronousTester` which provides the logic above by default.

leonardt: That would make sense, but there's an issue with backwards compatibility if we remove this init clock logic from the base tester

leonardt: (i.e. existing TBs that rely on the default clock initialization logic that don't use synchronous tester)

rsetaluri: I think I see the problem. Things like `LoopTester` need to inherit from `Tester` and not `TesterBase` to get all the implementations of normal functions. So we have to override `init_clock` explicitly somehow in `LoopTester` etc.

This seems fine to me in light of that.

PR #255, opened 2020-08-17
Add support for expect message

leonardt:


leonardt:
Previously, but magma made it an optional dependency so we have to install it explicitly now for some tests.

PR #258, opened 2020-09-03
Add Tester logic to find default clock

leonardt:
Traverses the interface and uses a clock port if there is only one.  If
there are multiple, we ask the user to choose one explicitly rather than
guessing.

### DIFF
--- a/fault/tester/base.py
+++ b/fault/tester/base.py
@@ -28,6 +28,10 @@ class TesterBase:
         self.actions = []
         if clock is not None and not isinstance(clock, m.Clock):
             raise TypeError(f"Expected clock port: {clock, type(clock)}")
+        else:
+            clock = self._find_default_clock(
+                self._circuit.interface.ports.values()
+            )
         self.clock = clock
         # Make sure the user has initialized the clock before stepping it
         # While verilator initializes the clock value to 0, this assumption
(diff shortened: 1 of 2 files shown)


fault/tester/base.py
```diff
@@ -41,6 +46,24 @@ def __init__(self, circuit: m.Circuit, clock: m.Clock = None,
...
 
+    def _find_default_clock(self, ports):
+        """Automatically finds a clock if there is only one"""
+        clock = None
+        for port in ports:
+            next_clock = None
+            if isinstance(port, m.Clock):
+                next_clock = port
+            elif isinstance(port, (m.Array, m.Tuple)):
+                nested_clock = self._find_default_clock(port)
```
rsetaluri, line 57: does this need to be `self._find_default_clock([port])`

leonardt: Nope, because it will iterate over the port (over it's children), otherwise we would infinitely loop on this array port

leonardt: I changed it to use port.ts for clarity (this is what it was doing implicitly)

rsetaluri: got it now, thanks.

rsetaluri (approved):

PR #263, opened 2020-09-23
add xcelium support

Kuree:
Since `irun` and `xrun` have the exact same commands, I just merge them together for better readability and easier maintenance.


sgherbst (approved):
Thanks!  LGTM

PR #266, opened 2020-10-01
Control actions should produce Tester based on current subclass

leonardt:
Before, using a control structure such as `tester._if` would produce a
StagedTester instance.  This is problematic when using a
SynchronousTester, since the produced tester doesn't inherit the
expected methods (e.g. advance_cycle).

Also fix unrelated magma error from missing `m.enable` cast (regression introduced in latest magma release)

### DIFF
--- a/fault/tester/control.py
+++ b/fault/tester/control.py
@@ -1,0 +1,58 @@
+import magma as m
+from typing import List
+
+
+class LoopIndex:
+    def __init__(self, name):
+        self.name = name
+
+    def __str__(self):
+        return self.name
+
+
+class NoClockInit:
+    """
+    Sub testers should not initialize the clock
+    """
+
+    def init_clock(self):
+        pass
+
+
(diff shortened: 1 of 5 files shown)


rsetaluri (approved):
Nice pattern. Was thinking this might be hard to do but this is elegant!

leonardt:
Yes, I had to muck around with various not so great solutions.  I was thinking having class variables would be the way to go, but you can't define class variable that references the class (since it doesn't exist yet) so that was a dead end.  The easy way would just be to call a factory method inside __init__, but that would be bad for performance since we'd have to do it every time an object is created, although cacheing could avoid some of this.  I ended up with this which is essentially monkey patching (danger!!) after class definition (since you need a reference to the class to create the children), but at least we can wrap it cleanly inside the decorator pattern and avoids runtime cost by only having to do it at definition time.

PR #269, opened 2020-10-14
Nested compile guards

leonardt:
Allows the user to provide a list of compile guards (useful if some properties should require two defines).  We can't use `&&` inside `ifdef` so we have to generate nested `ifdef` statements.

### DIFF
--- a/doc/property.md
+++ b/doc/property.md
@@ -65,6 +65,20 @@ This will wrap the generated property inside a macro.
 `endif
 ```
 
+For multiple guards, you can pass a list of strings, for example:
+```python
+f.assert_(..., compile_guard=["ASSERT_ON", "FORMAL_ON"], ...)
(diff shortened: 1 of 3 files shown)


rsetaluri:
actually we can do

```c++
#if defined(COND1) && defined(COND2)
...
#endif 
```

same for `||`, `!`. `#ifdef` is just shorthand.

See https://www.cs.auckland.ac.nz/references/unix/digital/AQTLTBTE/DOCU_078.HTM for full spec.

leonardt:
that looks like documentation for C code, is verilog the same?

rsetaluri:
oops, for some reason I thought this was C/C++ code for verilator. yeah came to the same conclusion as you for verilog.

rsetaluri (approved):

PR #270, opened 2020-10-16
Add pono example

leonardt:
Here's a little example of working with Pono to verify a SimpleALU.

I'm going to work on a `fault` frontend to generate this code, mainly the state machine generation can be abstracted using a poke sequence, and the property specification could use SMTBitVector.  This way we can then hook up the Lassen RTL (replace SimpleALU) with the Lassen SMTBitVector (replace the `spec_res`).  We can use the state machine generated by configure the Lassen core.

@makaimann can you check out this code and let me know if there's any place I can improve it? It's based on randomly browsing the examples, tests, header files, so I may have missed some useful functions/patterns.

I'll also need to setup Pono in the CI environments for this to work.


tests/test_pono.py
```diff
@@ -0,0 +1,175 @@
...
+
+    # State machine variable
+    test_state = rts.make_statevar('test_state', bvsort2)
+
+    # Starts at state 0
+    rts.constrain_init(solver.make_term(Equal, test_state,
+                                        solver.make_term(0, bvsort2)))
+
+    # sequential states 0 -> 1 -> 2 -> 3, then stays at 3
+    rts.assign_next(
```
makaimann, line 80: This could be:
```
rts.assign_next(test_state,
        rts.make_term(Ite,
                      solver.make_term(Equal, test_state, rts.make_term(3, bvsort2)),
                      rts.make_term(3, bvsort2),
                      rts.make_term(BVAdd, test_state, rts.make_term(1, bvsort2))))
```

But I don't think it makes much of a difference performance-wise -- just slightly less code. I will say that the way you wrote it is probably more realistic for the general case, i.e. if we get an arbitrary state machine, we might not be able to optimize it with a small encoding. Additionally, if it'd be helpful, I can add some kind of helper function to pono to generate this kind of encoding automatically. We'd just need to agree on an interface.

tests/test_pono.py
```diff
@@ -0,0 +1,175 @@
...
+        config_en_driver,
+        solver.make_term(
+            Ite,
+            solver.make_term(
+                Equal,
+                test_state,
+                solver.make_term(0, bvsort2)
+            ),
+            solver.make_term(True),
+            solver.make_term(
```
makaimann, line 133: I think this could just be `solver.make_term(False)` here. Although as mentioned above, this might be more realistic for an automatically generated state machine for driving the reset sequence.

On that note: you'll likely get better performance if you're able to just start the circuit from a concrete reset state. I.e. instead of encoding the reset sequence, just simulate the reset sequence, and assign any of the non-X values in the initial state for formal. That's assuming you're not verifying the reset sequence of course. Not sure how practical that is in general.

tests/test_pono.py
```diff
@@ -0,0 +1,175 @@
...
+                         solver.make_term(config_data, bvsort2))
+    )
+
+    # property imp_res matches spec_res when test_state == 3
+    prop = pono.Property(
+        rts, solver.make_term(Or, solver.make_term(Equal, imp_res, spec_res),
+                              solver.make_term(Distinct, test_state,
+                                               solver.make_term(3, bvsort2))))
+
+    # set solver options
```
makaimann, line 171: It'd actually be best to set the solver options above, right when you create the solver.

makaimann (approved):
Looks good to me! I left a few minor comments that you can consider. Regarding getting pono in Travis, I could make a new PyPi release. Unfortunately, we can't distribute it with mathsat, so you wouldn't be able to run interpolant-based model checking.

tests/test_pono.py
leonardt: Nice, yes that's more compact. I think for basic fault test cases they'll use this sequential/incrementing pattern, but once we add support for branches/loops it'll be a bit more complicated.  For an interface, perhaps some state machine object representation? (e.g. states, transitions) There are a few existing Python libraries for this, so perhaps we could leverage one of them as a representation and compile it down to Pono terms.  @phanrahan has used https://github.com/qntm/greenery with magma in the past and it seemed pretty nice (and simple)

tests/test_pono.py
leonardt: Hmm, that makes sense, it seems like combining simulation+formal in this way might be a cool technique to explore.  I'll see how we can support it and then we can run some experiments.  I'd imagine for complex systems this might be a good idea, particularly if we can "cache" the reset state based on the design (e.g. we can use the output verilog as a hash, and so long as it doesn't change, we don't need to recompute the reset state).  This would avoid having to redo the logic either in simulation or formal

leonardt: One thing I wasn't sure about, I originally had this code:
```diff
diff --git a/tests/test_pono.py b/tests/test_pono.py
index eb33a79..914eae4 100644
--- a/tests/test_pono.py
+++ b/tests/test_pono.py
@@ -105,7 +105,7 @@ def test_pono(config_data):
                 test_state,
                 solver.make_term(0, bvsort2)
             ),
-            solver.make_term(True),
+            solver.make_term(False),
             solver.make_term(
                 Ite,
                 solver.make_term(
@@ -113,7 +113,7 @@ def test_pono(config_data):
                     test_state,
                     solver.make_term(1, bvsort2)
                 ),
-                solver.make_term(False),
+                solver.make_term(True),
                 solver.make_term(
                     Ite,
                     solver.make_term(
```
Basically, the `config_en` was high a cycle later.  However when I do this, the tests fail.  Is there some relationships between states and cycle position? (e.g. does a state count as a half cycle or full cycle?) Or is there some delay required? (which is why moving up the time of the assertion of config en a state lets the result propagate?)

tests/test_pono.py
makaimann: Cool, that makes a lot of sense! I'll look into greenery.

tests/test_pono.py
makaimann: Ahh interesting. My guess is that the issue actually had to do with the property. Since it's not guarded by the value of `test_state` then if you have `config_en` go high a step later, then it can just violated the property in the initial state. You could try changing the `config_en` logic back but then make the property: `test_state >= 1 -> <original prop>`.

Regarding the cycle/state relationship, it depends on the encoding. Right now, the CoreIR encoder detects this automatically by default. So our "abstracted clock" encoding means that every state transition is a full clock cycle. This happens iff there's a single clock and no negedge behavior. Otherwise, each state transition is a "step" and the clock toggling has to be done manually: so normally you'd have a fastest clock that toggles every step and then possibly slower clocks on a different period. The CoreIR encoder also takes a boolean that can force it to abstract the clock (all clocks become synchronized and neg-edge behavior is dropped): https://github.com/upscale-project/pono/blob/157c4c83728b1d6926a5376f0a7a6b04ddf69c4b/frontends/coreir_encoder.h#L15. I haven't tested that feature much though, so let me know if you run into issues.

leonardt: Hmm, I was thinking that the `Distinct` portion of the property took care of that?
```
    prop = pono.Property(
         rts, solver.make_term(Or, solver.make_term(Equal, imp_res, spec_res),
                               solver.make_term(Distinct, test_state,
                                                solver.make_term(3, bvsort2))))
```
unless the initial state is undefined and that portion of the `or` returns false?

leonardt: (I was going for: either state is not 3 or the property must be true

makaimann: Oops, no you're right -- I just didn't look carefully. Hmm could you print the witness and send it to me? After getting `FALSE`, you can just do `print(rts.witness())`

leonardt: Here's the witness for when `config_data = 3` (doing bvudiv)
```
[{(bvadd a (bvneg b)): (_ bv1 16), spec_res.next: (_ bv2 16), (ite (= ((_ extract 0 0) config_reg$conf_reg$reg_P_inst0) (_ bv1 1)) (bvadd a (bvneg b)) (bvadd a b)): (_ bv3 16), config_reg$conf_reg$reg_P_inst0.next: (_ bv0 2), test_state.next: (_ bv1 2), config_reg$conf_reg$reg_P_inst0: (_ bv0 2), config_en_driver.next: false, CLK: false, config_en_driver: false, CLK.next: false, (bvmul a b): (_ bv2 16), (bvudiv a b): (_ bv2 16), (let ((.def_28 (= ((_ extract 0 0) config_reg$conf_reg$reg_P_inst0) (_ bv1 1)))) (ite (= ((_ extract 1 1) config_reg$conf_reg$reg_P_inst0) (_ bv1 1)) (ite .def_28 (bvudiv a b) (bvmul a b)) (ite .def_28 (bvadd a (bvneg b)) (bvadd a b)))): (_ bv3 16), (= ((_ extract 1 1) config_reg$conf_reg$reg_P_inst0) (_ bv1 1)): false, (ite (= ((_ extract 0 0) config_reg$conf_reg$reg_P_inst0) (_ bv1 1)) (bvudiv a b) (bvmul a b)): (_ bv2 16), config_en: false, (ite config_en config_data config_reg$conf_reg$reg_P_inst0): (_ bv0 2), imp_res: (_ bv0 16), b: (_ bv1 16), a: (_ bv2 16), test_state: (_ bv0 2), (bvadd a b): (_ bv3 16), spec_res: (_ bv0 16), (= ((_ extract 0 0) config_reg$conf_reg$reg_P_inst0) (_ bv1 1)): false, config_data: (_ bv3 2), imp_res.next: (_ bv3 16)}, {test_state.next: (_ bv2 2), spec_res.next: (_ bv1 16), imp_res.next: (_ bv9 16), (ite (= ((_ extract 0 0) config_reg$conf_reg$reg_P_inst0) (_ bv1 1)) (bvadd a (bvneg b)) (bvadd a b)): (_ bv9 16), CLK.next: false, CLK: false, config_en_driver.next: true, (bvmul a b): (_ bv20 16), config_reg$conf_reg$reg_P_inst0.next: (_ bv0 2), (bvudiv a b): (_ bv1 16), (bvadd a (bvneg b)): (_ bv1 16), (let ((.def_28 (= ((_ extract 0 0) config_reg$conf_reg$reg_P_inst0) (_ bv1 1)))) (ite (= ((_ extract 1 1) config_reg$conf_reg$reg_P_inst0) (_ bv1 1)) (ite .def_28 (bvudiv a b) (bvmul a b)) (ite .def_28 (bvadd a (bvneg b)) (bvadd a b)))): (_ bv9 16), (= ((_ extract 1 1) config_reg$conf_reg$reg_P_inst0) (_ bv1 1)): false, (ite config_en config_data config_reg$conf_reg$reg_P_inst0): (_ bv0 2), (ite (= ((_ extract 0 0) config_reg$conf_reg$reg_P_inst0) (_ bv1 1)) (bvudiv a b) (bvmul a b)): (_ bv20 16), config_en: false, imp_res: (_ bv3 16), b: (_ bv4 16), a: (_ bv5 16), config_en_driver: false, config_reg$conf_reg$reg_P_inst0: (_ bv0 2), (bvadd a b): (_ bv9 16), test_state: (_ bv1 2), config_data: (_ bv3 2), (= ((_ extract 0 0) config_reg$conf_reg$reg_P_inst0) (_ bv1 1)): false, spec_res: (_ bv2 16)}, {test_state.next: (_ bv3 2), spec_res.next: (_ bv65535 16), imp_res.next: (_ bv0 16), (ite (= ((_ extract 0 0) config_reg$conf_reg$reg_P_inst0) (_ bv1 1)) (bvadd a (bvneg b)) (bvadd a b)): (_ bv0 16), CLK.next: false, CLK: false, config_en_driver.next: false, (bvmul a b): (_ bv0 16), config_reg$conf_reg$reg_P_inst0.next: (_ bv3 2), (bvudiv a b): (_ bv65535 16), (bvadd a (bvneg b)): (_ bv0 16), (let ((.def_28 (= ((_ extract 0 0) config_reg$conf_reg$reg_P_inst0) (_ bv1 1)))) (ite (= ((_ extract 1 1) config_reg$conf_reg$reg_P_inst0) (_ bv1 1)) (ite .def_28 (bvudiv a b) (bvmul a b)) (ite .def_28 (bvadd a (bvneg b)) (bvadd a b)))): (_ bv0 16), (= ((_ extract 1 1) config_reg$conf_reg$reg_P_inst0) (_ bv1 1)): false, (ite config_en config_data config_reg$conf_reg$reg_P_inst0): (_ bv3 2), (ite (= ((_ extract 0 0) config_reg$conf_reg$reg_P_inst0) (_ bv1 1)) (bvudiv a b) (bvmul a b)): (_ bv0 16), config_en: true, imp_res: (_ bv9 16), b: (_ bv0 16), a: (_ bv0 16), config_en_driver: true, config_reg$conf_reg$reg_P_inst0: (_ bv0 2), (bvadd a b): (_ bv0 16), test_state: (_ bv2 2), config_data: (_ bv3 2), (= ((_ extract 0 0) config_reg$conf_reg$reg_P_inst0) (_ bv1 1)): false, spec_res: (_ bv1 16)}]
```

leonardt: I see `config_reg$conf_reg$reg_P_inst0: (_ bv0 2)` and `config_reg$conf_reg$reg_P_inst0: (_ bv0 2)` in the final state of the witness, so it looks like the register value hasn't been updated yet in the final state

leonardt: Hmm, perhaps because `config_en_driver` is a state variable, there's a 1 cycle delay for it to be updated, which in turn takes another cycle for `config_data` to be updated, so we need the extra cycle/state for the register value to be updated

makaimann: Something is strange with `config_en_driver` as well. It looks like it's toggling between values, but it should just be `true` in the first state and then `false` in the rest. I'll try to reproduce this locally.

makaimann: I think you're right. `config_en_driver` doesn't go high until the 3rd state because at state 2, `test_state==1`. Then it takes one more step for the `config_reg` to be updated, so the property can be violated exactly when `test_state` first becomes 3. You could add an extra bit to `test_state` and have it go to 4 before checking the property if you want to test that?

leonardt: Yea, that's essentially what moving the config_en True up a state does.  I'm looking into whether this is an issue for fault, because there are some "combinational" checks done in normal tests, whereas it seems like Pono requires us to only use state variables inside properties. We could just restrict properties at the fault level to be similarly stateful, but is there a way to write a property dependent on something like the next value of a state variable?

makaimann: Right, so properties in Pono are restricted to being over state variables. I know it's somewhat annoying because CoSA was more relaxed about it, but it's much more predictable in terms of behavior. The simplest way to get around that is just to create a property signal, e.g. a state variable `prop` that keeps track of whether the property is satisfied (this extends the trace by 1).

For example, say you want to check if `x.next == x`. Then you just create a variable `xnext_prop` that starts as `True` and then constrain it like: `rts.constrain_trans(solver.make_term(Iff, rts.next(xnext_prop), solver.make_term(Equal, xnext, x)`. And now your property is just `xnext_prop`. I'm planning to handle that automatically in `Property` for you soon (we already handle it automatically in the binary). I can make a PR for that today. I'm not sure I'll stick with the implementation, but the interface should stay the same at least.

Kuree:
Maybe try to release `smt-switch` and `pono` to PyPI so that the CI setup would be much easier? It will also make AHA flow happy.

makaimann:
That makes sense. I’ll make a new release later this week. However, I can’t distribute it with mathsat through PyPi. And mathsat is required for some algorithms, such as this one. There might be a way to dynamically load it though.

leonardt:
One other thing to note is how a coreir and pono wheel will interact.  Presumably the pono wheel would be statically linked with libcoreir?  This is probably problematic if we want to use both libraries at once (they would be using different versions of the library).  Perhaps for this flow it's not a problem since we could maybe just isolate pono for this task (use magma/coreir to generate a json, then switch to just pono/coreir to load the json into the different context).  Since coreir provides a wheel, could pono rely on the package for the dependency and avoid statically linking coreir?

makaimann:
Yeah that's a great point. I don't know much about that to be honest. It definitely seems possible that we could just depend on the `coreir` wheel and not link statically. At the C++ level, if it's dynamically linked then presumably it would all work out as long as it's in the load path.

Kuree:
As far as I know there isn't an easy way to bypass `auditwheel`'s dependency checking. One popular workaround is to use `dlopen` instead of of dynamically/static linking.

leonardt:
The issue is that the coreir wheel installs the library into the python package path, which is not part of the standard load path.   The `coreir` script shipped with the wheel wraps the binary in a script that adds the load path for the python package.  So pono can't assume that a version of coreir installed via pip will be available in the load path by default.  We could probably make pono aware of the wheel distribution and try to load it if it's present.  @makaimann if you can release a pono wheel I can look into making it work with the coreir wheel.  Even for now, we can manually add the pycoreir directly to the load path just for CI purposes, but shipping a wheel would make integration easier since I'm still having trouble getting it to build and install correctly in this environment

leonardt:
(FWIW it's fine locally, for some reason the github workflow environment is wonky and it's harder to debug)

makaimann:
Hey @leonardt, I made a PyPi release for Pono. I made a mistake and didn't add the correct requires field, so you'll need to separately install smt-switch, i.e. `pip install smt-switch && pip install pono`. I'll fix that in the next release. I'm not sure when I'll get to it, but at some point I'll look into using `dlopen` for the `coreir` library, and maybe also for the `msat` backend which we can't distribute but is required for several features.

leonardt:
@makaimann ok, looks like it's working, thanks for the help! I just needed to switch to btor since msat doesn't ship with the pypi package, but that's totally fine for now.

makaimann:
Great! Glad it's working. I'll try to improve it with future releases. In particular, it seems reasonable to dynamically load a mathsat backend so you can still use it even though we can't ship it with the pypi package.

PR #271, opened 2020-10-19
Add API to just generate TB file

leonardt:
Feature request from @kavyasreedhar to just generate a test bench (without running it)

### DIFF
--- a/fault/system_verilog_target.py
+++ b/fault/system_verilog_target.py
@@ -722,15 +722,19 @@ class SystemVerilogTarget(VerilogTarget):
         # return the string representing the system-verilog testbench
         return src
 
-    def run(self, actions, power_args=None):
+    def generate_test_bench(self, actions, power_args=None):
+        if self.ext_test_bench:
+            raise Exception(
+                "Cannot generate_test_bench when using external test bench")
         # set defaults
         power_args = power_args if power_args is not None else {}
+        return self.write_test_bench(actions=actions, power_args=power_args)
 
+    def run(self, actions, power_args=None):
         # assemble list of sources files
         vlog_srcs = []
         if not self.ext_test_bench:
-            tb_file = self.write_test_bench(actions=actions,
-                                            power_args=power_args)
+            tb_file = self.generate_test_bench(actions, power_args)
             vlog_srcs += [tb_file]
         if not self.ext_model_file:
             vlog_srcs += [self.verilog_file]
--- a/README.md
+++ b/README.md
@@ -330,3 +330,14 @@ tester.poke(circ.CLK, 0)
 tester.step()
 tester.print("%08x\n", circ.O)
 ```
+
+
+### Can I just generate a test bench without running it?
+Yes (currently only supported by the verilator target, please open an issue to
+request support for another target).  Here's an example:
+```
+# compile the tester
+tester.compile("verilator")
+# generate the test bench file (returns the name of the file)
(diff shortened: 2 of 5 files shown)


rsetaluri:
For system verilog target we have: https://github.com/leonardt/fault/blob/b634f16d145d6b2113e51e0d68b643e558c8b0f1/fault/system_verilog_target.py#L147-L148. Perhaps we could make them the same interface. Fine with `skip_run` flag or calling `generate_test_bench()` after `compile()`. If we can unify the interfaces then it would be great.

leonardt:
Hmm, I think part of the confusion was that `compile` didn't actually produce the test bench (I think this is because the verilator code path uses it to compile the verilator model (which can be reused across multiple test bench drivers), while `run` generates the test bench file.  `compile_and_run` with the `skip_run` flag was a hook to get the test bench without running it, but it's bit weird because the test bench is actually generate inside the` run` method. I think separating it into a different function makes the most sense for clarity, so we separate the compile, test bench generation, and run phases explicitly (we can keep the skip_run flag for backwards compatibility).

fault/system_verilog_target.py
```diff
@@ -722,15 +722,19 @@ def generate_code(self, actions, power_args):
         # return the string representing the system-verilog testbench
         return src
 
-    def run(self, actions, power_args=None):
+    def generate_test_bench(self, actions, power_args=None):
```
rsetaluri, line 725: looks like a different interface? i.e. one takes in actions one takes in target? It does seem a bit weird to have to pass in target into compile and then generate_test_bench.

What if you do

```python
tester.compile("target1", ...)
tester.generate_test_bench("target2", ...)
```

leonardt: If you haven't compiled `target2`, you'll get an error (https://github.com/leonardt/fault/pull/271/files#diff-39f4c94be89b8849a9043859f74ec7aef421fcb4ff22b3d08c629a82e89f993cR239-R244)
If you have compiled `target2` it will work fine.

The interface supports multiple compilation targets, although I don't think anyone is using this feature.  I think the main reason to separate `compile` from `run` (and now generate_test_bench) was that we wanted to support pre compiling the verilator model to save time in test runs, so some tests just call generate_test_bench or run.  If we didn't want to support multiple compilation targets for a test, we could then attach a specific target and not have to specify the targets for generate_test_bench/run

leonardt: Also the targets have a different interface than the tester because the tester passes it's actions through to the target (so the target API isn't really user facing, it's meant to be invoked via the tester API)

rsetaluri: Oops my mistake on target vs tester classes.

Got it. Would argue in general this pattern could be a little non-robust if you do weird stuff like more actions in between compile and run or something. Or call things in weird orders. But that's definitely "user error".

rsetaluri (approved):

leonardt:
Yea, in reviewing this code we may want to think of a better front-end API.  I think we can keep around the existing methods for backwards compatibility, but it's not very extensible to do `compile_and_run` and then try to add another step (e.g. do we add `compile_and_generate_test_bench_and_run`.  I think maybe we can devise a single front-end entry point method with options (then this avoids the user doing anything weird in between them).  However, there's still the use case of "precompiling" (I guess we could have a `skip_compile` option to skip a stage, one annoyance is that we already have that option for skipping the magma compile)

PR #272, opened 2020-10-22
Imm assert

leonardt:
Adds support for verilog style immediate assertion syntax

### DIFF
--- a/fault/assert_immediate.py
+++ b/fault/assert_immediate.py
@@ -1,8 +1,9 @@
 import magma as m
+from fault.property import Posedge
 
 
 def assert_immediate(cond, success_msg=None, failure_msg=None, severity="error",
-                     on="*"):
+                     on=None):
     """
     cond: m.Bit
     success_msg (optional): passed to $display on success
@@ -10,6 +11,8 @@ def assert_immediate(cond, success_msg=None, failure_msg=None, severity="error",
                             wrapped in quotes, integers are passed without
                             quotes (for $fatal))
     severity (optional): "error", "fatal", or "warning"
+    on (optional): If None, uses always @(*) sensitivity, otherwise something
+                   like f.posedge(io.CLK)
     """
     success_msg_str = ""
     if success_msg is not None:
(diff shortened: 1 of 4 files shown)


fault/assert_immediate.py
```diff
@@ -0,0 +1,38 @@
...
+            failure_msg = f"\"{failure_msg}\""
+        elif isinstance(failure_msg, int):
+            failure_msg = str(failure_msg)
+        else:
+            raise TypeError(f"Unexpected type for failure_msg={type(failure_msg)}")
+        failure_msg_str = f" else ${severity}({failure_msg});"
+    format_args = {}
+    if on is None:
+        on = "@(*)"
+    elif isinstance(on, Posedge):
```
rsetaluri, line 32: why only posedge?

rsetaluri (approved):
curious: what's the fundamental difference between immediate assert and "normal" assert?

fault/assert_immediate.py
leonardt: That's the only supported pattern right now even for concurrent properties, eventually we should replace this with something like `Event` and `Posedge` should subclass `Event`, but in practice most things are just `*` or `posedge(CLK)`, so we haven't needed to add support for other events yet

leonardt:
From http://staging.doulos.com/knowhow/systemverilog/systemverilog-tutorials/systemverilog-assertions-tutorial/
> Immediate assertions are procedural statements and are mainly used in simulation. An assertion is basically a statement that something must be true, similar to the if statement. The difference is that an if statement does not assert that an expression is true, it simply checks that it is true, e.g.:
TLDR; immediate assertions are like `if` statements that raise an error upon failure (or warning) and the block structure allows you to include other code (so you could run some extra logic to dump state, or something).

Concurrent assertions are built with sequences, so they're fundamentally different in terms of target syntax (immediate: behavioral code inside always block, concurrent: separate syntax based on sequences).  I think it's possible to capture the same behavior in both, but it would use different syntax.  I'd imagine immediate assertions are simpler for basic software style assertions you're probably familiar with (assert some boolean condition is true, print if message if fail).  Concurrent assertions become more useful when you want to write SVA style declarative properties (as opposed to more traditional design style structural/imperative code)

PR #273, opened 2020-10-22
Fix bug in f.sva and not operator

leonardt:
Tracking down a bug, this seems to work, but wasn't being tested


leonardt:
Ah, turns out the issue was inside the `f.sva` version

rsetaluri (approved):

PR #274, opened 2020-10-23
Add support for name parameter to assert immediate

leonardt:


fault/assert_immediate.py
```diff
@@ -33,6 +34,7 @@ def assert_immediate(cond, success_msg=None, failure_msg=None, severity="error",
         on = on.compile(format_args)
     else:
         raise TypeError(f"Unexpected type for on={type(on)}")
+    name_str = "" if name is None else f"{name}: "
     m.inline_verilog(
-        "always {on} assert ({cond}){success_msg_str}{failure_msg_str};",
+        "always {on} {name_str}assert ({cond}){success_msg_str}{failure_msg_str};",
```
rsetaluri, line 39: need space before `assert`?

tests/test_assert_immediate.py
```diff
@@ -46,11 +49,14 @@ class Foo(m.Circuit):
         assert severity == "warning"
     out, _ = capsys.readouterr()
     if failure_msg is not None:
+        print(out)
```
rsetaluri, line 52: nit: remove print

rsetaluri (approved):

fault/assert_immediate.py
leonardt: `name_str` includes a space after it if it's present, so no space here (in the case where it's empty string, we avoid thee extra space)

PR #275, opened 2020-10-23
Add support for cover properties

leonardt:
Uses the same interface as `assert_` (intuitively, assert ignores when a property passes, cover ignores when a property fails).

Updates the `coverage=True` option to use coverage setting `a` for ncsim (alias for `all` which enables coverage reporting for cover statements, and includes the `b` (block) setting which was used previously)

Also updates the system verilog target (ncsim) to dump assertion information at the end of the run.


rsetaluri (approved):
nice abstraction!

Issue #277, opened 2020-10-28
fork/join Proposal


Kuree:
## Proposed semantics
The fork process is designed to be similar to `Loop`:
```Python
process_1 = tester.fork()
process_2 = tester.fork()

process_1.poke(circ.I0, 1)
process_2.poke(circ.I1, 2)

process_1.expect(circ.O0, 3)
process_2.expect(circ.O1, 4)

tester.join(process_1, process_2)
```

### Join semantics
We should support these three join semantics via function calls:
- `join` -> `tester.join(*args)`
- `join_none` -> `tester.join_none(*args)`
- `join_any` -> `tester.join_any(*args)`

### Implementation details
We need to take care of two major backends, namely, SystemVerilog and Verilator. SystemVerilog is much easier to implement without any doubt, but Verilator is straightforward as well.

#### SystemVerilog
Upon `join` functions are called, we extract actions from each process and codegen them inside a `begin end` block, then emit proper keyword.

#### Verilator
Although Verilator does not have any fork/join construct, we have `pthread` and `std::thread`. The tricky part is how to synchronize the local timing control, should we support it. One way to solve this is block the local timing control by default and use a scheduler thread to unblock fork thread. `join_none`, `join_any`, and `join` can be easily implemented through `pthread` or `std::thread`.

## Questions
1. It is very common for a fork process to have local timing controls. For instance, if default clock is set, we can do `##1` to wait until next clock cycle, or simply `#1` to delay one time unit. Are we going to support timing controls in fault? That might requires some changes on the implementation of steps and clocks in fault.
2. The concept of fork/join inevitably introduces race conditions. Are we going to support `mutex` or `semaphore` semantics? `semaphore` is a native construct in SystemVerilog, and it is almost identical to that in `pthread`. From implementation perspective it is straightforward to add.

Please let me know if you have any suggestions or comments.

leonardt:
[...]

PR #278, opened 2020-10-29
Add fork to systemverilog target

Kuree:
## Introduction
This is a prototype implementation of `fork` presented in #277 

## Usage
```Python
tester = fault.Tester(dut)
proc1 = tester.fork("proc1")
proc2 = tester.fork("proc2")

for i, proc in enumerate([proc1, proc2]):
    proc.poke(tester.circuit.I[i], 1)
    proc.eval()
    proc.expect(tester.circuit.O[i], 1)

tester.join(proc1, proc2)
```
Name is required for each forked process so it's easy to identify and useful for the waveform debugging. The current implementation should also support tested fork, as long as `join()` is called at the process, e.g. `proc.join(proc3, proc4)`.

Here is generated SV:
```SystemVerilog
    initial begin
        fork
        begin : proc1
            I[0] <= 1'b1;
            #1;
            if (!(O[0] === 1'b1)) begin
                $error("Failed on action=0 checking port dut.O[0] with traceback /home/keyi/workspace/fault/tests/test_fork.py:24.  Expected %x, got %x.", 1'b1, O[0]);
            end
            fork
            begin : proc3
                #1;
            end
            join
        end
        begin : proc2
            I[1] <= 1'b1;
            #1;
            if (!(O[1] === 1'b1)) begin
                $error("Failed on action=0 checking port dut.O[1] with traceback /home/keyi/workspace/fault/tests/test_fork.py:24.  Expected %x, got %x.", 1'b1, O[1]);
            end
        end
        join
        #20 $finish;
    end

```

## How it works
- Several new actions are introduced to allow fork and join: `Fork` and `Join` class
- `JoinType` is introduced to handle different join, i.e. `join`, `join_any`, and `join_none`. Helper functions are added so user can directly call `tester.join_any()` and `tester.join_none()`.
- Actions are added to the main tester only after join is called.

## Limitations
- Need to implement semaphore to avoid race conditions. In some cases the race conditions can be statically determined, which could be an interesting research project itself.

### DIFF
--- a/fault/actions.py
+++ b/fault/actions.py
@@ -1,3 +1,4 @@
+import enum
 from abc import ABC, abstractmethod
 
 from .ms_types import (RealIn, RealOut, RealInOut,
@@ -247,6 +248,40 @@ class Loop(Action):
                    self.actions]
         return Loop(n_iter=self.n_iter, loop_var=self.loop_var,
                     actions=actions, count=self.count)
+
+
+class Fork(Action):
+    def __init__(self, name, actions):
+        self.name = name
+        self.actions = actions
+
+    def __str__(self):
+        return f"Fork({self.name}"
+
+    def retarget(self, new_circuit, clock):
+        actions = [action.retarget(new_circuit, clock) for action in
+                   self.actions]
+        return Fork(self.name, actions)
+
+
+class JoinType(enum.Enum):
+    Default = enum.auto()
+    None_ = enum.auto()
+    Any = enum.auto()
+
+
+class Join(Action):
+    def __init__(self, *processes, join_type=JoinType.Default):
+        self.join_type = join_type
+        self.processes = list(*processes)
+
+    def __str__(self):
+        return f"Join({self.join_type.name})"
+
+    def retarget(self, new_circuit, clock):
+        processes = [process.retarget(new_circuit, clock) for process in
+                     self.processes]
+        return Join(*processes, join_type=self.join_type)
 
 
 class FileOpen(Action):
--- a/fault/actions.py
+++ b/fault/actions.py
@@ -271,17 +271,17 @@ class JoinType(enum.Enum):
 
 
 class Join(Action):
-    def __init__(self, *processes, join_type=JoinType.Default):
+    def __init__(self, processes, join_type=JoinType.Default):
         self.join_type = join_type
-        self.processes = list(*processes)
+        self.processes = processes
 
     def __str__(self):
         return f"Join({self.join_type.name})"
 
     def retarget(self, new_circuit, clock):
-        processes = [process.retarget(new_circuit, clock) for process in
-                     self.processes]
-        return Join(*processes, join_type=self.join_type)
+        processes = (process.retarget(new_circuit, clock) for process in
+                     self.processes)
+        return Join(processes, join_type=self.join_type)
 
 
 class FileOpen(Action):
--- a/fault/system_verilog_target.py
+++ b/fault/system_verilog_target.py
@@ -413,6 +413,19 @@ class SystemVerilogTarget(VerilogTarget):
         self.add_decl('integer', action.loop_var, exist_ok=True)
         return super().make_loop(i, action)
 
+    def make_join(self, i, action):
+        code = ["fork"]
+        for p in action.processes:
+            code += self.make_block(i, None, None, p.actions, p.name)
+        # join
+        if action.join_type == actions.JoinType.Default:
+            code += ["join"]
+        elif action.join_type == actions.JoinType.None_:
+            code += ["join_none"]
+        elif action.join_type == actions.JoinType.Any:
+            code += ["join_any"]
+        return code
+
     def make_file_open(self, i, action):
         # make sure the file mode is supported
         if not is_valid_file_mode(action.file.mode):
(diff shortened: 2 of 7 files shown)


fault/system_verilog_target.py
```diff
@@ -413,6 +413,19 @@ def make_loop(self, i, action):
...
+        code = ["fork"]
+        for p in action.processes:
+            code += self.make_block(i, None, None, p.actions, p.name)
+        # join
+        if action.join_type == actions.JoinType.Default:
+            code += ["join"]
+        elif action.join_type == actions.JoinType.None_:
+            code += ["join_none"]
+        elif action.join_type == actions.JoinType.Any:
+            code += ["join_any"]
```
leonardt, line 426: We can add an internal assertion here just in case someone adds a new join_type and forgets to add support here in the target.
```suggestion
            code += ["join_any"]
 else:
     assert False, f"Unexpected joint_type: {action.join_type}"
```

fault/actions.py
```diff
@@ -249,6 +250,40 @@ def retarget(self, new_circuit, clock):
...
+class JoinType(enum.Enum):
+    Default = enum.auto()
+    None_ = enum.auto()
+    Any = enum.auto()
+
+
+class Join(Action):
+    def __init__(self, *processes, join_type=JoinType.Default):
+        self.join_type = join_type
+        self.processes = list(*processes)
```
leonardt, line 276: Does this need to be converted to a list? By default `*args` will be a tuple, so we can avoid the conversion unless we plan to use the mutability

fault/actions.py
```diff
@@ -249,6 +250,40 @@ def retarget(self, new_circuit, clock):
...
+
+    def __str__(self):
+        return f"Join({self.join_type.name})"
+
+    def retarget(self, new_circuit, clock):
+        processes = [process.retarget(new_circuit, clock) for process in
+                     self.processes]
+        return Join(*processes, join_type=self.join_type)
```
leonardt, line 284: I think we can just leave this as a generator expression rather than constructing the list data structure then unpacking it right away
```suggestion
        processes = (process.retarget(new_circuit, clock) for process in
                     self.processes)
        return Join(*processes, join_type=self.join_type)
```

leonardt: Alternatively we could change the `Join` object to take in a list or tuple rather than packing/unpacking the arguments

leonardt, line 284: Fixed the suggestion to properly include the multiple lines:

```suggestion
        processes = (process.retarget(new_circuit, clock) for process in
                     self.processes)
        return Join(*processes, join_type=self.join_type)
```

tests/test_fork.py
```diff
@@ -0,0 +1,31 @@
...
+        proc.eval()
+        proc.expect(tester.circuit.O[i], 1)
+
+    tester.join(proc1, proc2)
+
+    tester.compile_and_run(
+        target=target,
+        simulator=simulator,
+        tmp_dir=True
+    )
```
leonardt, line 31: LGTM, I wonder if we could add a test that requires the `fork/join` semantics to be correct for it to work?

Simple idea: Each thread sets an index (like this test does), master thread after join checks the entire output (all indices) are set.

Another idea: simple circuit that passes through ready to valid.  One thread sets ready after a random number of cycles, other thread loops until ready is high and sets valid, both join when ready/valid are high.

Kuree: I prefer the second idea. Is there a `wait` action in fault?

leonardt: There is a `wait_until_high` function (https://github.com/leonardt/fault/blob/master/fault/tester/staged_tester.py#L390) which I think does what you want? There's also `wait_on` which takes a generic condition.

Kuree: I tried `wait_on` and `wait_until_high` and both of them complains about clock not specified. SystemVerilog has a `wait` construct that blocks the current process execution until the signal is high, and that construct does not depends on clock, see LRM: 9.2. Here is a simple example of `wait`:

```SystemVerilog
module top;
logic a;
initial begin
    wait(a);
    $display("a is set @ %d", $time);
end
initial begin
    #10;
    a = 1;
end
endmodule
```

leonardt: Hmm, we can add support for `wait` in SV, I think that makes sense.

One problem though is how to map this concept to `verilator` which is a cycle accurate simulator? I guess one way to view it is `wait` will wait until the next simulator time step, so in verilator that would be every time `step` is called (normally this is defined w.r.t. to the clock in verilator, but in theory you could make `step` do something different if you were trying to model an analog circuit or something, but I'm not sure if this would work in the verilator model).  

Would the current `wait_on` implementation be equivalent to `wait` if we were looking at a purely synchronous design with one clock? Then one option is to stage `wait_on` in the frontend and compile based on the backend (so while loop in verilator, `wait` statement in system verilog)

Kuree: My understanding of Verilator is that `wait` semantics doesn't exists since the programming model assumes a single-thread driver code. After `eval()` is called, the state of circuit is always synchronized and you can just use `if` statement to test out inside a loop. Partially it is because Verilator is not an event-driven simulator, unlike commercial simulators. In my opinion, if there is no multi-threaded driver implementation in Verilator, we would never worry about `wait` semantics.

leonardt (approved):
New test looks good, I propose we merge this and then do a follow up PR with support for SV wait.  There's some other users who are interested in trying this out.

PR #282, opened 2020-11-05
Assert examples

leonardt:
Adds 2 more complex assertion examples to the documentation and tests.

Minor improvement to make them work:
* need to support expressions with property constructs (e.g. `value != f.past(value, 2)`).  For now, we piggy back on fault's expression language used for peek values.  In the future, we should consider supporting this in the downstream tools (e.g. we can treat `f.past(value, 2)` as a verification instance.  I think we can just represent these structures in the graph, but we probably want to group them together so they're code generated separate from the RTL (we don't want verification code to introduce RTL-like instances that may affect synthesis or cause the synthesis/verification code to have a mismatch).

### DIFF
--- a/doc/property.md
+++ b/doc/property.md
@@ -265,21 +265,24 @@ class Foo(m.Circuit):
     io = m.IO(a=m.In(m.Bits[8]), b=m.In(m.Bits[8]), c=m.In(m.Bits[8]),
               x=m.Out(m.Bits[8]), y=m.Out(m.Bits[8]))
     io += m.ClockIO(has_resetn=True)
+    # NOTE: Circuit code should appear above or else referring to output values
+    # (e.g. `io.y.value()` won't work)
     # sva syntax
     f.assert_(
         f.sva(f.not_(f.onehot(io.a)), "&&",
               io.b.reduce_or(), "&&",
-              io.x[0], "|=>", io.y != f.past(io.y, 2)
-        ),
+              io.x[0].value(), "|=>",
+              io.y.value() != f.past(io.y.value(), 2)),
         name="name_A",
         on=f.posedge(io.CLK),
         disable_iff=f.not_(io.RESETN)
     )
     # infix syntax
     f.assert_(
-        f.not_(f.onehot(io.a)) &
-        io.b.reduce_or() &
-        io.x[0] |f.implies| io.y != f.past(io.y, 2),
+        # Note parens matter!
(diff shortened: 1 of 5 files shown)


doc/property.md
```diff
@@ -246,3 +246,104 @@ def test_coverage(capsys):
...
+```python
+class Foo(m.Circuit):
+    io = m.IO(a=m.In(m.Bits[8]), b=m.In(m.Bits[8]), c=m.In(m.Bits[8]),
+              x=m.Out(m.Bits[8]), y=m.Out(m.Bits[8]))
+    io += m.ClockIO(has_resetn=True)
+    # NOTE: Circuit code should appear above or else referring to output values
+    # (e.g. `io.y.value()` won't work)
+    # sva syntax
+    f.assert_(
+        f.sva(f.not_(f.onehot(io.a)), "&&",
```
rsetaluri, line 272: I've probably been out of the loop on is but: it seems like there are many different ways to do "logic" on signals. `&&` is represented as a string, `onehot` seems to be a verification (fault) specific function, and `|` (or-reduce) seems to be traditional magma logic. Curious why it's this way.

rsetaluri (approved):

doc/property.md
leonardt: "&&" is slightly different than "&" when it comes to precedence, but it's possible to do the equivalent logic with magma & rather than the "&&" sva syntax, I just chose to do it this way since it's based off an existing SVA example and it was closest to the original code (otherwise you have to make sure you get the parenthesis right to get the equivalent logic, which is what the infix example shows).  `onehot` is a fault function because there's no way to pass arguments to the sva operator strings, so we have to add Python level support for functions (the alternative would be doing something like `, "$onehot(", io.a, ")", ...` which seems undesirable, especially since we need the functions for the infix syntax anyways.

leonardt: (&& is similar to `and` in python (w.r.t. &), but as you know, we can't override that behavior)

rsetaluri: Got it. So it seems like constructing the asserts out of magma circuits is not really plausible.

leonardt: It would require downstream support.  E.g. we could have $onehot be a primitive that we instance (wire up in the inputs/outputs), but coreir would need to be aware of how to code generate the primitive (function call instead of structural).  coreir could add a similar `&&` operator, but here the issue is really a frontend/syntax problem (we could try a functional syntax like `f.and(a, b, c)`)

PR #283, opened 2020-11-13
Add support for abs, max, min expressions

leonardt:
Needed for floating point tests (off by one errors mean we should be checking `abs(value - expected) <= 1`


Kuree (approved):

leonardt:
@Kuree this doesn't add support for max/min in ncsim since that requires the macro implementation, let me know if you need this and I will add it, should be simple.

PR #284, opened 2020-11-13
Defines, parameters, and expanded real-number support for Verilator

sgherbst:
## Summary
This PR adds support for defining macros and module parameters with Verilator, using arguments called ``defines`` and ``parameters``, which have the same meaning as in SystemVerilogTester.  It also makes it possible to write ``expect`` statements involving real numbers with Verilator.

## Details
1. The ``defines`` argument is a dictionary whose ``var``/``value`` entries map to Verilator ``-D<var>[=<value>]`` command-line arguments.
2. The ``parameters`` argument is a dictionary whose ``name``/``value`` entries map to Verilator ``-G<name>[=<value>]`` command-line arguments.
3. For ``expect`` statements, the Verilator target previously cast all received and expected values to ``unsigned int``, but now won't do this when the port in question carries a real value.
4. Implemented ``above`` and ``below`` optional arguments for ``expect`` with the Verilator target, allowing for range comparisons in addition to equality comparisons.  This is important for writing tests about real numbers, where equality checks are generally to be avoided.
4. Added ``ext_model_file`` argument for Verilator target: when provided, it sets ``skip_compile`` to ``True`` and is used instead of the Verilog file name determined by the parent class.  This makes it possible to simulate a non-generated DUT that is in an arbitrary location.
5. Updated ``test_def_vlog.py``, ``test_param_vlog.py``, and ``test_real_val.py`` to exercise these new features in the Verilator target.
6. Minor formatting fixes in unrelated files, to pass ``pycodestyle``.  (Not sure how they passed before?)
7. Bumped package version in anticipation of a PyPI release after merging.


leonardt (approved):
LGTM

PR #286, opened 2020-11-20
add support to poke and expect var

Kuree:
Known issue: 
https://github.com/leonardt/fault/blob/ff27d9466d25d98b277f953992bb0317cfe2e0be/fault/verilator_target.py#L596-L602

Not sure how to get sign from that. Maybe there is another class I should check?

### DIFF
--- a/fault/verilator_target.py
+++ b/fault/verilator_target.py
@@ -343,6 +343,8 @@ if (!({cond})) {{
                 # to use top->v instead of top->{circuit_name}
                 name += f"{prefix}->"
             name += action.port.verilator_path
+        elif isinstance(action.port, actions.Var):
+            name = action.port.name
         else:
             name = verilator_name(action.port.name)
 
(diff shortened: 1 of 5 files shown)


fault/verilator_target.py
```diff
@@ -589,9 +594,17 @@ def make_file_read(self, i, action):
         ])
 
     def make_var(self, i, action):
-        if isinstance(action._type, AbstractBitVectorMeta) and \
-                action._type.size == 32:
-            return [f"unsigned int {action.name};"]
+        if isinstance(action._type, AbstractBitVectorMeta):
```
leonardt, line 597: I think something like this would work:
```python
if issubclass(action._type, SIntVector):
    <signed code>
elif issubclass(action._type, BitVector):  # works for both BitVector and UIntVector
    <unsigned code>
raise NotImplementedError
```

leonardt (approved):
LGTM

Issue #280, opened 2020-11-03
Extremely low performance for large test vectors


Kuree:
When I did an exhaustive test on floating points, I noticed performance issue with fault generated testbench. The test vectors size is 0x500 * 0x500 = 0x190000 ~ 1 million data points. Here is the profiling result:
![image](https://user-images.githubusercontent.com/6099149/98027249-4052e480-1dc1-11eb-9392-12cbbc9ae832.png)

The entire runtime takes 22k seconds, about 6 hours. As a comparison, complete exhaustive test using SystemVerilog + DPI only takes 10 hours to finish (4 billion test points). So the performance gap is orders of magnitude (2000x).

Notice that the RTL simulation takes about 12k seconds to run. This is due to the sheer size of testbench code generated, which is 408M.

I think there are several places where it can be improved:
1. `hwtypes` computation. The native C implementation is about 100x faster. I looked at the actual float implementation and it is very inefficient:
```Python
exp = 1 - bias
s = ['-0.' if sign[0] else '0.', mantissa.binary_string()]
s.append('e')
s.append(str(exp))
return cls(gmpy2.mpfr(''.join(s),cls.mantissa_size+1, 2))
```
It is converting to and back from native gmp object using string. I think native conversion is available?
2. Getting frame information, particular filename and line number. I had the same performance issue before and was able to resolve it by implementing the same logic in native Python C API. It is included in `kratos` package. I can give you a pointer on how to use it, if you're interested. On my benchmark it is about 100-500x faster than doing the same thing in Python.

The real questions is whether the inefficient test bench is an intrinsic problem in fault. My understanding is that all the expect values have to be known during compilation time, which makes the testbench unscalable when there are lots of test vectors. One way to solve that is to allow simulator call python code during simulation. I have a prototype working: https://github.com/Kuree/kratos-dpi/blob/master/tests/test_function.py#L18-L28
It works with both Verilator and commercial simulators.

Please let me know what you think.

leonardt:
[...]

PR #289, opened 2020-11-21
Add native python functional model

Kuree:
## Usage example:
DUT:
```Python
class dut(m.Circuit):
    io = m.IO(A=m.In(m.Bits[4]), B=m.In(m.Bits[4]), O=m.Out(m.Bits[4]))
    io.O @= io.A + io.B
```

Functional model (used for demonstrate class constructor and method call):
```Python
class Model:
    @sv(b=DataType.UByte)
    def __init__(self, b):
        self.b = b
 
   @sv(a=DataType.UByte)
    def add(self, a):
        return a + self.b
```

Test code:
```Python

a_value = 1
tester = fault.Tester(dut)
tester.poke(tester.circuit.A, a_value)

model = tester.Var("model", Model)
tester.poke(model, tester.make_call(Model, a_value))

# start testing. using a loop
loop = tester.loop(10)
loop.poke(tester.circuit.B, loop.index)
loop.eval()
loop.expect(tester.circuit.O, tester.make_call(model.add, loop.index))
```

## Nee functions/class introduced
1. Add `make_call` to tester class
2. Add `CallExpression`  to `expression`
3. Extend `Var` to support `var.foo`, where `var` is declared `tester.Var`, and `foo` is a function name. Compile time check to ensure correctness

## Code refactored
1. Refactor template the support start and exit code region. This can be used for simulator environment setup and teardown.

## Simulator tested
- Xcelium/Incisive
- VCS
- Verilator

## Discussion
This should solve #280 fundamentally by allowing arbitrary Python code execution inside a simulator. It also enables runtime model check with values unknown during compilation time.

Most of the heavy lifting is done via the `pysv` package. Currently there are some limitation on how the code should rewritten. Notably how the class and functions import are handled. If there is enough use case for that I can spend some time to solve that issue.

### Roadmap
- I will continue work on bringing the support for arrays. Ideally it should not introduce any changes to fault.
- I also talked to @sgherbst  and he suggested that it might also be possible to do something similar with SPICE simulator. I will look into that.

### DIFF
--- a/fault/system_verilog_target.py
+++ b/fault/system_verilog_target.py
@@ -226,7 +226,7 @@ class SystemVerilogTarget(VerilogTarget):
                           DeprecationWarning)
             self.dump_waveforms = dump_vcd
         self.no_warning = no_warning
-        self.imports = []
+        self.imports = set()
         self.declarations = {}  # dictionary keyed by signal name
         self.assigns = {}  # dictionary keyed by LHS name
         self.sim_env = sim_env if sim_env is not None else {}
@@ -534,8 +534,7 @@ class SystemVerilogTarget(VerilogTarget):
             return [f'if (!({expr_str})) $error("{expr_str} failed");']
 
     def make_call(self, i, action: actions.Call):
-        if "pysv" not in self.imports:
-            self.imports.append("pysv")
+        self.imports.add("pysv")
         return super().make_call(i, action)
 
     def make_expect(self, i, action):
@@ -704,7 +703,7 @@ class SystemVerilogTarget(VerilogTarget):
         port_list = f',\n{2*self.TAB}'.join(port_list)
 
         if len(self.pysv_funcs) > 0:
-            self.imports.append("pysv")
+            self.imports.add("pysv")
 
         # build up the body of the initial block
         initial_body = []
@@ -813,7 +812,7 @@ class SystemVerilogTarget(VerilogTarget):
             vlog_srcs = [os.path.relpath(pkg_file, self.directory)] + vlog_srcs
 
         # generate simulator commands
-        if self.simulator == 'ncsim' or self.simulator == "incisive" or self.simulator == "xcelium":
+        if self.simulator in {'ncsim', "incisive", "xcelium"}:
             # Compile and run simulation
             cmd_file = self.write_cadence_tcl()
             if self.simulator == "xcelium":
--- a/fault/system_verilog_target.py
+++ b/fault/system_verilog_target.py
@@ -868,7 +868,8 @@ class SystemVerilogTarget(VerilogTarget):
         # run the simulation binary (if applicable)
         if bin_cmd is not None:
             subprocess_run(bin_cmd, cwd=self.directory, env=self.sim_env,
-                           err_str=bin_err_str, disp_type=self.disp_type)
+                           err_str=bin_err_str, disp_type=self.disp_type,
+                           shell=True)
 
         # post-process GetValue actions
         self.post_process_get_value_actions(actions)
--- a/fault/system_verilog_target.py
+++ b/fault/system_verilog_target.py
@@ -868,8 +868,7 @@ class SystemVerilogTarget(VerilogTarget):
         # run the simulation binary (if applicable)
         if bin_cmd is not None:
             subprocess_run(bin_cmd, cwd=self.directory, env=self.sim_env,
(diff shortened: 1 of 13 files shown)


Kuree:
Oddly the tests passed for fpga on buildkite but not the normal one. It seems like conda has some weird python system path setup. I will take a look into it. One workaround is to use `venv` instead of conda on `cadence7`, which I've successfully tested.

leonardt:
The conda env is setup with the following commands:
```
conda create -p ./conda_env -m -y pip numpy python=3.7.3
export PATH=$PWD/conda_env/bin:$PATH
```
so it creates a local environment in the current working directory.  I'm guessing we need to configure pysv to use the correspond library load path? Is there a way we can point it to a custom directory, e.g. with an environment variable? I guess this is similar to how `venv` works with a special directory, so is there anything special done to configure that?

Could it be that `pysv` is building/linking with the wrong python? How does it find out which one to use? Since we're only editing `PATH`, maybe we're not setting the right environment variable like PYTHONHOME?

leonardt:
Relevant part of the workflow: https://github.com/leonardt/fault/blob/master/.buildkite/pipeline.yml#L12-L13

Kuree:
For libraries, pysv uses the `sys.path` of whatever Python it is running with and then have the compiled runtime load that, see [here](https://pysv.readthedocs.io/advanced/internal.html#foreign-definitions).

As for linking, it uses pybind11's built in capability to detect currently python executable and libraries. The error seems to look like library loading issue.

Kuree:
@leonardt I tried to prune unnecessary imports from pysv's side to avoid switching `sys.path`, still can't reproduce the error.

Here is my working directory on cadence7: `/tmp/fault-pysv/`

```
pytest tests/test_pysv.py -v
```

Conda is created at `/tmp/fault-pysv/conda_env` using the exact same command.

Tried with several Linux distro with conda and still can't reproduce the error. Would be nice if all the tests are containerized and easy to reproduce.

fault/system_verilog_target.py
```diff
@@ -681,6 +703,9 @@ def generate_code(self, actions, power_args):
             port_list.extend(result)
         port_list = f',\n{2*self.TAB}'.join(port_list)
 
+        if len(self.pysv_funcs) > 0:
+            self.imports.append("pysv")
```
leonardt, line 707: Looking at 538, which guards against adding pysv more than once, could this add a pysv import more than once? Perhaps we could just simply make `imports` a set?

fault/system_verilog_target.py
```diff
@@ -766,19 +803,25 @@ def run(self, actions, power_args=None):
...
-        if self.simulator == 'ncsim' or self.simulator == "incisive":
-            # Compile and run simulation
-            cmd_file = self.write_cadence_tcl()
-            sim_cmd = self.ncsim_cmd(sources=vlog_srcs, cmd_file=cmd_file)
-            sim_err_str = None
-            # Skip "bin_cmd"
-            bin_cmd = None
-            bin_err_str = None
-        elif self.simulator == 'xcelium':
+        if self.simulator == 'ncsim' or self.simulator == "incisive" or self.simulator == "xcelium":
```
leonardt, line 816:
```suggestion
        if self.simulator in {"ncsim", "incisive", "xcelium"}:
```

fault/verilator_target.py
```diff
@@ -730,6 +751,14 @@ def run(self, actions, verilator_includes=None, num_tests=0,
         else:
             env = None
 
+        # codegen the c++ binding if needed
+        if len(self.pysv_funcs) > 0:
+            header = os.path.join(self.directory, "pysv.hh")
+            pysv.generate_cxx_binding(self.pysv_funcs, filename=header)
+            lib_path = pysv.compile_lib(self.pysv_funcs, os.path.join(self.directory, "obj_dir"))
+            if env is None:
+                env = {"LD_LIBRARY_PATH": os.path.realpath(os.path.dirname(lib_path))}
```
leonardt, line 760: Should we have an else case here where we either insert LD_LIBRARY_PATH (if it doesn't exit) or append to it (if it exists)?

leonardt (approved):
LGTM, finally got CI passing, some minor comments to address before merging

fault/system_verilog_target.py
Kuree: I was worried about non-deterministic code generation result. It seems like none of the tests are directly comparing the code output so I think using set would be better.

PR #290, opened 2020-12-04
Raise an error when trying to set an invalid port

cdonovick:


fault/wrapper.py
```diff
@@ -19,29 +19,42 @@ def __setattr__(self, attr, value):
...
+        except AttributeError:
+            init_done = False
+
+        if init_done:
             if attr in self.circuit.interface.ports.keys():
                 return PortWrapper(self.circuit.interface.ports[attr], self)
             elif attr in self.instance_map:
                 return InstanceWrapper(self.instance_map[attr], self)
             else:
                 object.__getattribute__(self, attr)
```
cdonovick, line 56: Just noticing this is probably an error right?  I expect it should be `return object.__getattribute__(self, attr)`.

leonardt: yes it should be

leonardt:
CI is failing on an unrelated pysmt-install issue.  We're transition CI to a docker based flow in this branch (https://github.com/leonardt/fault/pull/289) so I may wait to get that working rather than fixing it here and having to deal with the conflicts.  For now I'll just disable the msat install (this was used for the Cosa test and we're deprecating that for Pono anyways and the Pono tests uses boo lector so it's unnecessary)

leonardt (approved):

PR #291, opened 2021-01-12
Simplify assert immediate to use initial block

leonardt:
In verilog, assertions are "behavioral" statements so they need to be
contained within a block such as always or initial (compared to
structural statements such as module instancing or assignment).  Before,
we used always @(*), however initial is preferred for some tool
compatability an the effect should be the same (since it's an immediate
assertion, it doesn't really make sense to have it sensitive to anything
other than *, so we can remove the code related to other events).


leonardt:
Looks like the buildkite is failing related to a license problem on the CI server

PR #294, opened 2021-01-14
Move raise AssertionError to its own function

cdonovick:


fault/subprocess_run.py
```diff
@@ -75,6 +75,11 @@ def error_detected(text, err_str):
         raise Exception(f'Invalid err_str: {err_str}.')
 
 
+def _throw(args, cwd, env, disp_type, err_str,
+           chk_ret_code, shell, use_fault_cfg):
+    raise AssertionError()
```
leonardt, line 80: Can we add a short comment noting why we do this?

leonardt (approved):
LGTM

fault/subprocess_run.py
cdonovick: Sure, will do

cdonovick: I added the comment at the call site as opposed to here so we don't need to read the comment on every failed test.

PR #298, opened 2021-02-09
Add basic support for pysv monitor

leonardt:
A monitor is a pysv class that defines an `observe` function that is
invoked every time advance_cycle is called.

The arguments to the observe functions refer to circuit ports by name
(in the future we could add support for internal signals).

### DIFF
--- a/tests/test_pysv.py
+++ b/tests/test_pysv.py
@@ -132,6 +132,8 @@ def test_monitor(target, simulator):
     def test(circuit, enable):
         tester = fault.SynchronousTester(circuit)
         monitor = tester.Var("monitor", Monitor)
+        # TODO: Need clock to start at 1 for proper semantics
+        tester.poke(circuit.CLK, 1)
         tester.poke(monitor, tester.make_call_expr(Monitor))
         tester.attach_monitor(monitor)
         tester.poke(circuit.CE, enable)
--- a/tests/test_pysv.py
+++ b/tests/test_pysv.py
@@ -1,3 +1,4 @@
+import pytest
 from pysv import sv, DataType
 from pysv.util import clear_imports
 import fault
@@ -13,12 +14,15 @@ def pytest_generate_tests(metafunc):
 
 class dut(m.Circuit):
     io = m.IO(A=m.In(m.Bits[4]), B=m.In(m.Bits[4]), O=m.Out(m.Bits[4]))
+    io += m.ClockIO()
+    io.CLK.unused()
     io.O @= io.A + io.B
 
 
 def run_tester(tester, target, simulator):
     kwargs = {
         "target": target,
+        # "disp_type": "realtime",
         "simulator": simulator,
         "tmp_dir": False
     }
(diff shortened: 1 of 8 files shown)


tests/test_pysv.py
```diff
@@ -97,5 +101,52 @@ def incr(self, amount):
...
+def test_monitor(target, simulator):
+    class DelayedDUT(m.Circuit):
+        io = m.IO(A=m.In(m.Bits[4]), B=m.In(m.Bits[4]), O=m.Out(m.Bits[4]))
+        io += m.ClockIO(has_enable=True)
+        io.O @= m.Register(m.Bits[4], has_enable=True)()(dut()(io.A, io.B),
+                                                         CE=io.CE)
+
+    # TODO: Using a subclass breaks pysv
+    # class Monitor(fault.PysvMonitor):
+    class Monitor:
```
leonardt, line 113: @Kuree is there anything specific I need to do to make pysv work with generic python subclasses? When I use the abstract base class, I get an error
```
>       clear_imports(Monitor)

tests/test_pysv.py:128:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../pysv/pysv/util.py:92: in clear_imports
    func_defs = make_unique_func_defs(func_defs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

func_defs = [<class 'tests.test_pysv.test_monitor.<locals>.Monitor'>]

    def make_unique_func_defs(func_defs):
        # first remove any functions that's class method
        result = []
        classes = []
        for func_def in func_defs:
            if type(func_def) == type:
                if func_def not in classes:
                    result.append(func_def)
                    classes.append(func_def)
            else:
                if hasattr(func_def, "func_def"):
                    func_def = func_def.func_def
>               if func_def.parent_class is not None:
E               AttributeError: type object 'Monitor' has no attribute 'parent_class'

../pysv/pysv/util.py:74: AttributeError
```

tests/test_pysv.py
```diff
@@ -97,5 +101,52 @@ def incr(self, amount):
...
+            self.value = None
+
+        @sv()
+        def observe(self, A, B, O):
+            print(A, B, self.value)
+            if self.value is not None:
+                assert O == self.value, f"{O} != {self.value}"
+            # TODO: Referencing bitvector here doesn't work, so just manually
+            # mask for now
+            # self.value = BitVector[4](A) + BitVector[4](B)
```
leonardt, line 125: Also, in the pysv github readme, there's an example using numpy inside the class, I tried this with hwtypes BitVector and it didn't work, is there something I need to do to enable library support?

```
<STDERR>
terminate called after throwing an instance of 'pybind11::error_already_set'
  what():  NameError: name 'BitVector' is not defined

At:
  <string>(12): observe

</STDERR>
```

tests/test_pysv.py
Kuree: I never tried subclass so there might be some assumptions in the code that break the usage. I will do some test locally.

tests/test_pysv.py
Kuree: That sounds like a bug. I will try to reproduce it locally.

tests/test_pysv.py
Kuree: It works with normal subclass but not metaclass. I will implement additional check to catch metaclass inheritance.

tests/test_pysv.py
Kuree: Okay got it figured out:
1. Unfortunately importing foreign class is not currently supported since it requires more complex import mechanism. I've added that as an issue and will work on that once I get more free time: https://github.com/Kuree/pysv/issues/11. One workaround is to use the module.
2. The imports are cleared before the codegen. I think I added it for other tests to avoid importing modules because I couldn't get it working with the test environment. It should be removed now that the test environment is working.

Here is what I did to make it working locally on my machine:
```diff
diff --git a/tests/test_pysv.py b/tests/test_pysv.py
index 94803b4..20d2e61 100644
--- a/tests/test_pysv.py
+++ b/tests/test_pysv.py
@@ -5,6 +5,7 @@ import fault
 import magma as m
 from hwtypes import BitVector
 from .common import pytest_sim_params
+import hwtypes
 
 
 def pytest_generate_tests(metafunc):
@@ -123,12 +124,10 @@ def test_monitor(target, simulator):
                 assert O == self.value, f"{O} != {self.value}"
             # TODO: Referencing bitvector here doesn't work, so just manually
             # mask for now
-            # self.value = BitVector[4](A) + BitVector[4](B)
-            self.value = (A + B) & ((1 << 4) - 1)
+            self.value = hwtypes.BitVector[4](A) + hwtypes.BitVector[4](B)
+            # self.value = (A + B) & ((1 << 4) - 1)
             print(f"next value {self.value}")
 
-    clear_imports(Monitor)
-
     def test(circuit, enable):
         tester = fault.SynchronousTester(circuit)
         monitor = tester.Var("monitor", Monitor)

```

tests/test_pysv.py
Kuree: Should be fixed with the latest release.

tests/test_pysv.py
Kuree: Ignore the deleted comment if you've already seen it. I spoke too soon and forgot that I actually resolve the import limitation some time ago.

All you need is to remove `clear_imports`. I think I used it when I was hacking around the test environment to make the library load properly. Now that the testing environment is working, `clear_imports` should be removed.

Here is what I did locally to pass the test:
```diff
diff --git a/tests/test_pysv.py b/tests/test_pysv.py
index 94803b4..8e6b7f8 100644
--- a/tests/test_pysv.py
+++ b/tests/test_pysv.py
@@ -123,12 +123,10 @@ def test_monitor(target, simulator):
                 assert O == self.value, f"{O} != {self.value}"
             # TODO: Referencing bitvector here doesn't work, so just manually
             # mask for now
-            # self.value = BitVector[4](A) + BitVector[4](B)
-            self.value = (A + B) & ((1 << 4) - 1)
+            self.value = BitVector[4](A) + BitVector[4](B)
+            # self.value = (A + B) & ((1 << 4) - 1)
             print(f"next value {self.value}")
 
-    clear_imports(Monitor)
-
     def test(circuit, enable):
         tester = fault.SynchronousTester(circuit)
         monitor = tester.Var("monitor", Monitor)

```

Kuree (approved):

PR #302, opened 2021-03-18
Add support for compiling UnknownValue

leonardt:
This allows checking that a value is X, which could be useful for
verifying initialization behavior.

### DIFF
--- a/fault/system_verilog_target.py
+++ b/fault/system_verilog_target.py
@@ -15,6 +15,7 @@ from fault.wrapper import PortWrapper
 from fault.subprocess_run import subprocess_run
 import fault
 import fault.expression as expression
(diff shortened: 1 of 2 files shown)


rsetaluri (approved):

leonardt:
@Kuree looks like the kratos test is failing: https://buildkite.com/stanford-aha/fault/builds/2125#ca49cdc2-d42d-406a-873a-119de6bc5d51 was there a recent change that requires a code update here?

Kuree:
There hasn't been any update for past couple months so it's unlikely related. That being said I think we should remove/disable the test since I plan to migrate everything to hgdb once magma supports source-level debugging.

PR #311, opened 2021-07-26
Update fault inline wire prefix for immediate assertions

leonardt:


fault/assert_immediate.py
```diff
@@ -4,7 +4,8 @@
 
 
 def _make_assert(type_, cond, success_msg=None, failure_msg=None,
-                 severity="error", name=None, compile_guard=None, delay=False):
+                 severity="error", name=None, compile_guard=None, delay=False,
+                 inline_wire_prefix="_FAULT_ASSERT_WIRE_"):
```
rsetaluri, line 8: might be worth using a pattern like

```python
def _make_assert(..., inline_wire_prefix=None):
    if inline_wire_prefix = None:
        inline_wire_prefix = "_FAULT_ASSERT_WIRE_"
```

 then in the public functions `assert_immediate`, etc, you can just do:

```python
def assert_immediate(..., inline_wire_prefix=None):
    ...
    _make_assert(..., inline_wire_prefix=inline_wire_prefix)
```

This way the default logic is only in one place. (If you wanted to change it, you would only have to change it in one place.)

rsetaluri (approved):

PR #314, opened 2021-09-21
Add ReadyValidTester for I/O sequences

leonardt:
For an initial implementation we can actually create a synthesizable
test bench to feed input sequences and check output sequences.  This is
likely to become an issue when scaling to large sequences (since we have
to generate each sequence element as a constant fed into a mux).  To
handle this, I think we can change the sequence mux to read data from a
file (or a memory that has its contents loaded from a file).


leonardt:
Depends on https://github.com/phanrahan/magma/pull/992

fault/tester/ready_valid_tester.py
```diff
@@ -0,0 +1,52 @@
...
+            for key, value in sequences.items():
+                port = getattr(inst, key)
+                n = len(value)
+                # Counter to advance through sequence
+                count = m.Register(m.UInt[(n - 1).bit_length()],
+                                   has_enable=True)()
+                count.I @= count.O + 1
+
+                # Current sequence elem chosen by counter
+                curr = m.mux(value, count.O)
```
rsetaluri, line 21: Isn't the idea to use a fifo here?

fault/tester/ready_valid_tester.py
```diff
@@ -0,0 +1,52 @@
+import magma as m
+from fault.tester.staged_tester import Tester
+from fault.assert_immediate import assert_immediate
+
+
+class ReadyValidTester:
+    def __init__(self, circuit: m.Circuit, sequences: dict):
```
rsetaluri, line 7: by the way the type annotation should be `circuit: m.circuit.CircuitKind`

fault/tester/ready_valid_tester.py
```diff
@@ -0,0 +1,52 @@
+import magma as m
+from fault.tester.staged_tester import Tester
+from fault.assert_immediate import assert_immediate
+
+
+class ReadyValidTester:
```
rsetaluri, line 6: It's a bit odd to call this class `ReadyValidTester` and not inherit from `BaseTester`. Ostensibly the "right" thing to do would be to inherit from `BaseTester` and then raise exceptions on all the peek/poke/etc type methods? But I understand that's a bit non-robust because everytime a method is added there, we would have to nullify it there.

Is there any reason to structure it as a class? I think it makes more sense to do the following:
* Define a separate function `wrap_with_sequence` (prob a better name than this...) that just returns a new synthesizable circuit which is the Wrapper you defined here.
* Then have one function like
```python
def run_ready_valid_test(ckt: m.DefineCircuitKind, sequences: Mapping, synthesizable: bool = True):
    if synthesizable:
        wrapped = wrap_with_sequence(ckt, sequences)
        tester = Tester(wrapped)
        tester.wait_until_high(wrapped.done)
        tester.compile_and_run(*args, **kwargs)
        return
    # some logic to run in non-synthesizable mode, maybe a For-Loop tester?
```

rsetaluri (approved):

fault/tester/ready_valid_tester.py
leonardt: I'm not sure what a FIFO would add to this? We'd still need to drive/monitor the input/output sequences with this logic, we could stick a fifo in between the drivers/monitors but not sure how that would help (maybe useful for adding random stall logic)

leonardt: i.e. the FIFO has a ready/valid interface, so rather than plugging into the interface directly, we plug into the fifo in between them, not sure what this adds besides maybe allowing us to fill the fifo while the DUT is stalled?

leonardt:
I think we may be able to us the while loop tester to implement a non-synthesizable version since we need to wait until ready/valid handshakes, we could also use a for loop with a while loop inside to stall iterations, but we need some strategy for serializing the input/output sequences, we could try just having them as literals in the generated code, but using files might be a better longterm strategy (assuming we can compress things better into a binary representation for file IO)

leonardt:
Thought: how can we allow the user to debug issues with these tests? E.g. we can add a debug flag that prints out the sequences/handshakes (so perhaps we can find the one that it stalls on), is there someway we can inject user defined instrumentation code that might be useful? We can also just wait for user feedback to determine useful debugging features.

PR #320, opened 2022-05-18
Add option to SV target to generate packed arrays

rsetaluri:


tests/test_system_verilog_target.py
```diff
@@ -105,6 +107,26 @@ class X(m.Circuit):
...
+        io.O @= io.I
+
+    fault.Tester(_Foo).compile_and_run(
+        target="system-verilog",
+        simulator="vcs",
+        skip_compile=True,
+        skip_run=True,
+        use_packed_arrays=use_packed_arrays,
+    )
+    assert filecmp.cmp(f"gold/{name_}_tb.sv", f"build/{name_}_tb.sv")
```
leonardt, line 127: Can we add a test that double checks that at least pokes/peeks work with packed arrays from a functionality perspective?  There are some array tests in tests/test_tester/test_core.py that we could parametrize by the use_packed flag

leonardt (changes requested):

tests/test_system_verilog_target.py
leonardt: Ideally we'd test all the actions but I think poking/peeking would be a minimum

rsetaluri: @leonardt added peeks and pokes to the tests. Figure we want to check the generated testbench in the case that we can't run a SV simulator.

rsetaluri: also we can't use iverilog because it doesn't support nd arrays anyway

PR #326, opened 2023-08-01
Add broken tests

cdonovick:


cdonovick:
Due to some sort of caching issue I can't test `BrokenProtocol` and `FixedProtocol` at the same time.  Not sure how to resolve.

leonardt:
Is the cache issue at the magma/python level compile? Or the verilator test?

cdonovick:
> Is the cache issue at the magma/python level compile? Or the verilator test?

Sorry I should have provided more detail.  If I add `FixedProtocol` to the list of `BoxT` I get an attribute error about `BrokenProtocol` having no `debug_name` in the `FixedProtocol` cases (this is the error I get in the `BrokenProtocol` cases).  If I run only `FixedProtocol` it has no errors.  If I set `BoxT = [FixedProtocol, BrokenProtocol]` I get no errors.  So some sort of Python level caching issue.

leonardt:
I think a broader fix for this issue is to have `fault` unpack protocol values at the entrypoint to any fault code (i.e. the fault internals just operate on the underlying magma values).  I added the required unpacking for these tests, but I suspect there are (many) other places we may need to the `get_magma_value` logic.  However, this is likely much simpler to do at the API boundary rather than finding all the places internally where we need to call `get_magma_value`

